PR #929, opened 2024-11-13
Change/gat 4915: Multiple downloads on dataset landing page

spco:
## Describe your changes
A single endpoint serves 3 files' contents, depending on the `download_type` parameter.

Supports https://github.com/HDRUK/gateway-web-2/pull/881 FE

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-4915

### DIFF
--- a/app/Http/Controllers/Api/V1/DatasetController.php
+++ b/app/Http/Controllers/Api/V1/DatasetController.php
@@ -1313,12 +1313,77 @@ class DatasetController extends Controller
                     // var_dump($result['metadata']['metadata']);
                     // Add CSV headers
                     fputcsv($handle, $headerRow);
+                    $result = $result['metadata']['metadata'];
                     $rows = [
-                        ['Dataset', 'Name', $result['metadata']['metadata']['summary']['title']],
-                        ['Dataset', 'Gateway URL', $result['metadata']['metadata']['required']['revisions'][0]['url']], //?
-                        ['Dataset', 'Dataset Type', $result['metadata']['metadata']['summary']['datasetType']], //sort out delimiter
-                        ['Dataset', 'Dataset Sub-type', $result['metadata']['metadata']['summary']['datasetSubType']], //sort out delimiter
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -1186,6 +1187,223 @@ function () use ($results) {
...
+                    }
+                }
+
+                // Close the output stream
+                fclose($handle);
+            }
+        );
+
+        $response->headers->set('Content-Type', 'text/csv');
+        if ($download_type === 'structural') {
```
JamieByrneHDRUK, line 1395: Its really picky i know... but i feel like this could just be a conditional for the file name, rather than have have to set `$response->headers->set` each time just set something like $fileName based on the condition and use `$response->headers->set` once

spco: Fair comment

loki-sinclair-hdruk (approved):

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -1186,6 +1187,223 @@ function () use ($results) {
...
+        $dataset = Dataset::where('id', '=', $id)->first();
+
+        $result = $dataset->latestVersion()['metadata']['metadata'];
+
+        $response = new StreamedResponse(
+            function () use ($result, $download_type) {
+                // Open output stream
+                $handle = fopen('php://output', 'w');
+
+                if ($download_type === 'structural') {
```
loki-sinclair-hdruk, line 1254: Low prio, but I do insist upon some form of case sanity check. I usually, wrap the incoming variable in `strtolower` - just in case.

spco: done

app/Http/Controllers/Api/V1/DatasetController.php
spco: done

JamieByrneHDRUK (approved):

PR #933, opened 2024-11-18
GAT-5290: Collections permissions - split into 2 controllers

spco:
## Describe your changes
Implements a rewrite of Collections controller and route logic. The design is that Collections can be owned in one of two ways:
1. by a team, in which case the creating user is saved in `CollectionHasUser` as `CREATOR`, but all permissions checks are via team permissions
2. by an individual, in which case the creating user is saved in `CollectionHasUser` as `CREATOR`, and they may add collaborators via the `collaborators` field, in which case the supplied users are saved in `CollectionHasUser` as `COLLABORATOR`s.

This PR does the following to fulfil this design:
- New V2 API boilerplate added.
- Collections owned by **teams** are handled by `V1\CollectionController` with routes of the form `api/v1/teams/{teamId}/collections/{id}`. Their permissions are checked by `CheckAccessMiddleware` (which checks permissions of the user on the team in the route `{teamId}`) _and_ by `checkAccess()` Trait which checks the permissions of the user on the team that currently owns the collection.
- Collections owned by **individuals** are handled by `V2\CollectionController`, with routes in the form `api/v2/collections/{id}`. Their permissions are checked by `checkAccessCollaborators()` Trait - no middleware check is performed.
- Tests of all V1 and V2 endpoints.
  - Including new functionality in the `Authorization` Trait to easily create 2 non-admin users. This makes it easy to test access by other individuals without either having to be admin.
- Addition of new `Requests`, and modification of existing `Requests` to remove the use of `team_id` and `user_id` parameters, neither of which is now supported.

Update:
- Also split Requests into a `V2` folder, and explicitly ban the owning user of a team-owned Collection from editing it via the personal-Collection endpoints.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-5290 - this is BE to support the FE to come

None

No


routes/api.v2.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use Illuminate\Support\Facades\Route;
+
+Route::get('/test', function () {
+    return Response::json([
+        'message' => 'lorem ipsum dolor sit amet, consectetur adip',
+    ]);
+});
```
dnhdruk: you can remove

dnhdruk (commented):
for tests: don't mix tests for v2 with tests for v1. if v2 has been created, move everything to v2, not just partially.

app/Http/Requests/Collection/CreateCollection.php
```diff
@@ -98,10 +98,6 @@ public function rules(): array
                 'string',
                 'distinct',
             ],
-            'user_id' => [
-                'integer',
-                'exists:users,id',
-            ],
```
dnhdruk, line 104: dont mix v1 with v2: please create another form request.

spco: Ok. I'm aware that Requests don't have a V1 in the filepath - would you just split into `app/Http/Requests/V2/Collection/`, leaving the V1 in the parent folder `app/Http/Requests/Collection`? At least in this PR, I don't want to move all existing to a `V1` folder - though I'm happy to raise a later PR to that effect if it's required.

app/Http/Controllers/Api/V1/CollectionController.php
```diff
@@ -442,12 +442,23 @@ public function show(GetCollection $request, int $id): JsonResponse
 
     /**
      * @OA\Post(
```
JamieByrneHDRUK, line 446: isn't this going to break the existing route? the frontend is using this post endoint for `createCollection`

app/Http/Requests/Collection/CreateTeamCollection.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+namespace App\Http\Requests\Collection;
+
+use App\Http\Requests\BaseFormRequest;
+
+class CreateTeamCollection extends BaseFormRequest
```
dnhdruk, line 7: I don't see it used

app/Http/Requests/Collection/CreateCollection.php
dnhdruk: it's a good idea but it will be a lot to change, maybe just one folder, v2, where everything will be created for version 2

app/Http/Controllers/Api/V1/CollectionController.php
spco: It is - FE needs updating with an in-the-works PR at the same time to make use of the new route. It's not ideal, but the existing functionality is broken anyway so it's not (IMHO) a killer.

JamieByrneHDRUK: Ahhhh i seee

app/Http/Requests/Collection/CreateTeamCollection.php
spco: `CreateTeamCollection` is now used in V1 CollectionController@store. https://github.com/HDRUK/gateway-api-2/pull/933/files#diff-f2929124531d9eb3a0892f088aadeb6649990aa0148a62ed51ea69c685ac6c34R506

dnhdruk: work with v2 folder more

app/Http/Controllers/Api/V1/CollectionController.php
```diff
@@ -653,7 +670,7 @@ public function store(CreateCollection $request): JsonResponse
      *                   @OA\Property(property="datasets", type="array", example="[]", @OA\Items()),
      *                   @OA\Property(property="dur", type="array", example="[]", @OA\Items()),
      *                   @OA\Property(property="publications", type="array", example="[]", @OA\Items()),
-     *                   @OA\Property(property="users", type="array", example="[]", @OA\Items()),
+     *                   @OA\Property(property="collaborators", type="array", example="[]", @OA\Items()),
```
BSnelling: The code doesn't look like it handles collaborators, should this be removed from the swagger docs?

app/Http/Controllers/Api/V1/CollectionController.php
```diff
@@ -847,9 +867,8 @@ public function update(UpdateCollection $request, int $id): JsonResponse
      *                   @OA\Property(property="datasets", type="array", example="[]", @OA\Items()),
      *                   @OA\Property(property="dur", type="array", example="[]", @OA\Items()),
      *                   @OA\Property(property="publications", type="array", example="[]", @OA\Items()),
-     *                   @OA\Property(property="users", type="array", example="[]", @OA\Items()),
      *                   @OA\Property(property="applications", type="array", example="[]", @OA\Items()),
-     *                   @OA\Property(property="team", type="array", example="{}", @OA\Items()),
+     *                   @OA\Property(property="collaborators", type="array", example="[]", @OA\Items()),
```
BSnelling: Same as above, the collaborators handling has been removed from the code.

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -0,0 +1,1328 @@
...
+     * )
+     */
+    public function update(UpdateCollection $request, int $id): JsonResponse
+    {
+        $input = $request->all();
+        $jwtUser = $input['jwt_user'] ?? [];
+
+        // Allow access only to collaborators
+        $collHasUsers = CollectionHasUser::where(['collection_id' => $id])->select(['user_id'])->get()->toArray();
+        $this->checkAccessCollaborators($input, array_column($collHasUsers, 'user_id'));
```
BSnelling, line 275: Checking my understanding, if a collection is created using the `teams/{teamId}/collections` endpoint can it still be updated using this endpoint by the user listed as the `CREATOR`?

app/Http/Controllers/Api/V1/CollectionController.php
spco: This is only in the response, so the controller doesn't need to handle it itself. But you've highlighted to me that we _do_ actually return `users` rather than `collaborators` in the response. For simplicity, I'll revert to `users` in the docs rather than further complicating things by changing that model function too. The same is true below too.

app/Http/Controllers/Api/V2/CollectionController.php
spco: Yes, that's correct.

spco: Though that does raise a problem. If created for a team, then PUT via the V2 endpoint, it'll remove the `team_id` field, rendering it unavailable to the other users in the team. And the FE is not going to have a field to enter a `team_id` for the "personal collection" interface. We have 2 options 
- either in the V2 we explicitly copy across the `team_id` (we don't allow it as in input field, but we read it and use it when editing)
- or we disallow the user editing via this route. That would require the `checkAccessCollaborators` function to check for the presence of `team_id`, or we can put a simpler guard at the beginning of the V2 endpoints that if `team_id` is not null, we fail.

Thoughts @BSnelling ?

spco:
@BSnelling @dnhdruk re-reviews would be greatly appreciated now I've addressed all comments. thanks!

BSnelling (approved):

dnhdruk (commented):
tests are a mix of v1 and v2.

spco:
> tests are a mix of v1 and v2.

What do you mean by this? CollectionTest.php is entirely testing the V2 collections endpoints, while CollectionTeamTest.php is entirely testing the V1 collections endpoints.

dnhdruk (approved):

PR #937, opened 2024-11-20
adds new route for teams/{id}/datasets, specifically to trim the payl…

loki-sinclair-hdruk:
…oad for datasets administration list

## Screenshots (if relevant)
image

## Describe your changes
The datasets administration page (list) under Teams, suffers from pulling back the entire payload, which, for BHF, causes a server error (because it's massive). This provides the bare minimum this page needs, and now loads the BHF datasets in 128ms.

None.

No.

### DIFF
--- a/app/Http/Controllers/Api/V1/TeamController.php
+++ b/app/Http/Controllers/Api/V1/TeamController.php
@@ -1190,10 +1190,18 @@ class TeamController extends Controller
                 ->paginate($perPage, ['*'], 'page');
 
             foreach ($datasets as &$d) {
-                $d->latestMetadata['metadata'] = $this->trimDatasets($d->latestMetadata['metadata'], [
+                $miniMetadata = $this->trimDatasets($d->latestMetadata['metadata'], [
                     'summary',
                     'required',
                 ]);
+
+                // latestMetadata is a relation and cannot be assigned at this
+                // level, safely. So, unset all forms of metadata on the object
+                // and overwrite with out minimal version
+                unset($d['latest_metadata']);
+                unset($d['latestMetadata']);
+
+                $d['latest_metadata'] = $miniMetadata;
             }
 
             return response()->json([
--- a/app/Http/Controllers/Api/V1/TeamController.php
+++ b/app/Http/Controllers/Api/V1/TeamController.php
@@ -32,6 +32,7 @@ use App\Http\Traits\GetValueByPossibleKeys;
 use App\Http\Traits\IndexElastic;
 use App\Http\Traits\TeamTransformation;
 use App\Http\Traits\RequestTransformation;
+use App\Http\Traits\TrimPayload;
 
 use MetadataManagementController as MMC;
 
@@ -41,6 +42,7 @@ class TeamController extends Controller
     use IndexElastic;
     use TeamTransformation;
     use RequestTransformation;
+    use TrimPayload;
 
     private $datasets = [];
     private $durs = [];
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/V1/TeamController.php
```diff
@@ -1074,6 +1076,135 @@ public function destroy(DeleteTeam $request, int $teamId): JsonResponse
...
+                    $request->has('withTrashed'),
+                    function ($query) {
+                        return $query->withTrashed();
+                    }
+                )
+                // LS - This is how it should be done, other places we currently
+                // use both deleted_at and ARCHIVED as a status field. deleted_at
+                // should denote a record has _actually_ been deleted, and therefore
+                // hidden.
+                ->when($filterStatus, function ($query) use ($filterStatus) {
```
spco, line 1134: As per your comments, this is a change, and will need equivalent logic replicated in a _lot_ of places in the codebase. Currently the main way to archive an entity is to use the `DELETE` endpoint in the controller. And the use of `withTrashed` in various places will need reviewing etc. We rely in many places on `Dataset::where()` _not_ returning archived Datasets etc.

dnhdruk: I think we need to rely on `status` field not on `deleted_at` field.

loki-sinclair-hdruk: Yep, and as I've just said in slack, the way we're doing it now, is very far from correct. In traditional usage, `deleted_at` tells Laravel internal whether to bring the record back in the first place. This is partly why versioning didn't work. Too many variables to contend with.

spco: Agreed. I'd just be interested to know the migration plan to the new way of doing things (not now)! This will also require data migrations in prod to 'undelete' entities that should now be 'archived' only. I support it, just gulping at the changes that are required.

loki-sinclair-hdruk: Anything is possible, given the time ;)

dnhdruk (approved):

app/Http/Controllers/Api/V1/TeamController.php
loki-sinclair-hdruk: In theory, it's not difficult to determine actual state, but agreed it's a larger problem for sure. That said, easier to combat now, before we get too far ahead. Right now, we _could_ blanket nullify `deleted_at` where status is archived, and handle spot requests when they're found to be incorrect.

This new endpoint is isolated enough though, that it shouldn't cause too much trouble.

app/Http/Controllers/Api/V1/TeamController.php
```diff
@@ -1074,6 +1076,143 @@ public function destroy(DeleteTeam $request, int $teamId): JsonResponse
...
+
+            $tmp = explode(':', $sort);
+            $sortField = $tmp[0];
+            $sortDirection = array_key_exists('1', $tmp) ? $tmp[1] : 'asc';
+
+            $sortOnMetadata = str_starts_with($sortField, 'metadata.');
+            $allFields = collect(Dataset::first())->keys()->toArray();
+            if (!$sortOnMetadata && count($allFields) > 0 && !in_array($sortField, $allFields)) {
+                return response()->json([
+                    'message' => '\"' . $sortField . '\" is not a valid fiedl to sort on',
```
BSnelling: Typo:
```suggestion
                    'message' => '\"' . $sortField . '\" is not a valid field to sort on',
```

app/Http/Controllers/Api/V1/TeamController.php
```diff
@@ -1074,6 +1076,143 @@ public function destroy(DeleteTeam $request, int $teamId): JsonResponse
         }
     }
 
+    public function datasets(Request $request, int $teamId): JsonResponse
```
BSnelling, line 1079: Could you add swagger annotations?

BSnelling (approved):
LGTM, approving with minor comments.

app/Http/Controllers/Api/V1/TeamController.php
loki-sinclair-hdruk: This is purely internal, at present at least.

PR #938, opened 2024-11-20
fix: (GAT-5762) - Search should only show active collections

JamieByrneHDRUK:
## Describe your changes
Changed search controller to only show active collections
## Issue ticket link
https://hdruk.atlassian.net/jira/software/c/projects/GAT/boards/51?assignee=712020%3A05fbbf0f-9386-469a-a212-84cfbf9cc587&selectedIssue=GAT-5762

### DIFF
--- a/app/Http/Controllers/Api/V1/SearchController.php
+++ b/app/Http/Controllers/Api/V1/SearchController.php
@@ -683,7 +683,7 @@ class SearchController extends Controller
                 $matchedIds[] = $d['_id'];
             }
 
-            $collectionModels = Collection::whereIn('id', $matchedIds)->get();
+            $collectionModels = Collection::whereIn('id', $matchedIds)->where('status', 'ACTIVE')->get();
 
             foreach ($collectionArray as $i => $collection) {
                 $foundFlag = false;


spco (dismissed):

spco (commented):
Actually, does this address the ticket correctly? The bug ticket says a draft collection doesn't display in search, but does get shown as linked to a Tool, and it's then navigable from there. This fix would just hide a draft collection from search results, and the video in the ticket shows they're already hidden there? (but elastic won't return the draft collection since it's not indexed anyway, so this fix as it stands is probably not required anyway)

app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -683,7 +683,7 @@ public function collections(Request $request): JsonResponse
                 $matchedIds[] = $d['_id'];
             }
 
-            $collectionModels = Collection::whereIn('id', $matchedIds)->get();
+            $collectionModels = Collection::whereIn('id', $matchedIds)->where('status', 'ACTIVE')->get();
```
dnhdruk, line 686: I think we are using some fields in the collections table. I think we could make a statement for only certain fields vs. all fields.

dnhdruk (approved):

PR #945, opened 2024-11-25
Migrations for custom questions in qb

BSnelling:
## Describe your changes

Adds `qb_question_has_team` table so that questions can available to more than one team. Removes `team_id` column from `question_bank_questions` table.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-3977

Yes: `php artisan migrate`.  

### DIFF
--- a/database/migrations/2024_11_21_100036_create_qb_has_team_table.php
+++ b/database/migrations/2024_11_21_100036_create_qb_has_team_table.php
@@ -1,0 +1,56 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('qb_question_has_team', function (Blueprint $table) {
+            $table->id();
+            $table->bigInteger('team_id')->unsigned();
+            $table->bigInteger('qb_question_id')->unsigned();
+
+            $table->foreign('team_id')->references('id')->on('teams');
+            $table->foreign('qb_question_id')->references('id')->on('question_bank_questions');
+
+            $table->unique(['team_id', 'qb_question_id']);
+            $table->index('team_id');
+            $table->index('qb_question_id');
+        });
+
+        // Fill with existing data?
+        // Nulls mean all teams have access to the question
+        $questions = DB::select('select id, team_id from question_bank_questions');
+        // ??? all teams or select where is_question_bank is true
+        $qb_teams = DB::table('teams')->select('id')->pluck('id');
+
+        foreach ($questions as $question) {
+            if (is_null($question->team_id)) {
+                foreach ($qb_teams as $team) {
+                    DB::insert(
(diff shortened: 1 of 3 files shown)


dnhdruk (approved):

database/migrations/2024_11_21_100036_create_qb_has_team_table.php
```diff
@@ -0,0 +1,56 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('qb_question_has_team', function (Blueprint $table) {
```
calmacx, line 13: `question_bank_question_has_team` would be more consistent with the name of the existing tables

database/migrations/2024_11_21_100036_create_qb_has_team_table.php
```diff
@@ -0,0 +1,56 @@
...
+
+            $table->foreign('team_id')->references('id')->on('teams');
+            $table->foreign('qb_question_id')->references('id')->on('question_bank_questions');
+
+            $table->unique(['team_id', 'qb_question_id']);
+            $table->index('team_id');
+            $table->index('qb_question_id');
+        });
+
+        // Fill with existing data?
```
calmacx, line 26: shouldnt need to do this - I never implemented functionality for teams to own a question - though that was the long term plan

The original idea here was that if the `team_id` was null then it was a _global_ question that can be used by any team. If instead it had a team_id then it's a custom question that could only be used by that team... i'm guessing the scope changed to allow teams to share questions now

calmacx (approved):

database/migrations/2024_11_21_100036_create_qb_has_team_table.php
BSnelling: I agree but SQL complained about the index name being too long when I named the table that.

database/migrations/2024_11_21_100036_create_qb_has_team_table.php
BSnelling: Yes the ticket requires one or more teams to have access to a custom question. And for some questions to be available to all teams. 

I believe the only data in the qb table at the moment are questions belonging to all teams, so this migration adds rows to `qb_question_has_team` to reflect that.  Does that make sense for now?

calmacx: Yeah it does. Just seems a pain to manage and track adding/cascading when teams are created/deleted.

You'll see on the question bank management page that there are tabs for "active", "custom" and "archived". It was never implemented but the idea was that that would allow the superadmin to see and select all default questions and be able to manage the custom questions teams have created. There was to be a team question management page too.

Does this not also make it difficult to distinguish between a global and custom question?

BSnelling: Yes it would make that difficult. I didn't know about that requirement.  

Sounds like there needs to be an `is_global` column in the `question_bank_questions` table.

PR #952, opened 2024-12-03
Fix datasetsubtype filter

spco:
## Describe your changes
Change the current `datasetSubType` filter name to `dataSubType` to match the response from search-service.

## Issue ticket link
Fix for https://hdruk.atlassian.net/browse/GAT-5268, supporting https://hdruk.atlassian.net/browse/GAT-5369

Run `php artisan app:update-datasubtype-filter`

### DIFF
--- a/app/Console/Commands/UpdateDataSubTypeFilter.php
+++ b/app/Console/Commands/UpdateDataSubTypeFilter.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\Filter;
+use Illuminate\Console\Command;
+
+class UpdateDataSubTypeFilter extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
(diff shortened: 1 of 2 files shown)


dnhdruk (approved):

BSnelling (approved):
Apologies if this has been explained before but what is the reasoning behind updating the database via console commands rather than by migrations?

spco:
> Apologies if this has been explained before but what is the reasoning behind updating the database via console commands rather than by migrations?

I'm with you on this one - I think migrations have real benefits over commands, especially in local dev - far easier in local dev to run `migrate` than work out which commands are needed.

In the from-new case, the filters seeder is sufficient. In production, either command or migration would do the trick, but Dan objects to using migrations for this because it's a data migration not schema migration.

@BSnelling @dnhdruk are you happy if I _also_ add a migrations file? Then in local dev, developers don't need to work out which commands to run when. In prod, we can run the command or the migration, whichever is easier?

Or is that just unhelpful duplication that will cause more confusion than it saves?

PR #967, opened 2024-12-10
Feature/gat 5782

spco:
## Describe your changes
Adds nested question functionality to DAM.

Also renames `QuestionBankVersion::question_parent_id` to `question_id` to reduce confusion between the terminology used between nested questions (parent-child) and a `QuestionBankVersion` and its associated `QuestionBank` question.

~~This will be followed be a further PR that updates the QuestionBankSeeder to parse the migrated questions into this nested form. Keeping that out of this one for sanity's sake.~~
Update: I have also updated the Seeder just to keep it in one place

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-5782

Yes - 2 migrations, plus rerun the QuestionBankSeeder: `php artisan db:seed --class=QuestionBankSeeder` ⚠️ note that this _will drop all contents from the questions tables_ (`dar_sections`, `qb_question_has_team`, `question_bank_version_has_child_version`, `question_bank_versions`, `question_bank_questions`), so should only be run once, before any users make any changes to questions.

### DIFF
--- a/app/Http/Controllers/Api/V1/QuestionBankController.php
+++ b/app/Http/Controllers/Api/V1/QuestionBankController.php
@@ -273,84 +273,9 @@ class QuestionBankController extends Controller
                 'version' => 1,
             ]);
 
-            if (isset($input['team_id'])) {
-                foreach ($input['team_id'] as $t) {
-                    QuestionHasTeam::create([
-                        'qb_question_id' => $question->id,
-                        'team_id' => $t,
-                    ]);
-                }
-            } else {
-                $allTeams = Team::all()->select('id')->pluck('id');
-                foreach ($allTeams as $t) {
-                    QuestionHasTeam::create([
-                        'qb_question_id' => $question->id,
-                        'team_id' => $t,
-                    ]);
-                }
-            }
+            $this->updateQuestionHasTeams($question, $input);
 
-            // Don't allow children to also have children, and only allow certain parent types to have children
-            if (!($input['is_child'] ?? false)
-                && isset($input['children'])
-                && in_array($input['field']['component'], ['RadioGroup', 'CheckboxGroup', 'Autocomplete'])) {
-                // Create all children questions and question versions as required.
-                // All must by design have the same version number as the parent - parents and children move versions in lockstep
-                if (isset($input['children'])) {
-                    foreach ($input['children'] as $childListCondition => $childList) {
-                        foreach ($childList as $child) {
-                            $childQuestion = QuestionBank::create([
-                                'section_id' => $input['section_id'],
-                                'user_id' => $input['user_id'] ?? $jwtUser['id'],
-                                'force_required' => $child['force_required'],
-                                'allow_guidance_override' => $child['allow_guidance_override'],
-                                'locked' => $child['locked'] ?? false,
-                                'archived' => $child['archived'] ?? false,
-                                'archived_date' => ($child['archived'] ?? false) ? Carbon::now() : null,
-                                'is_child' => true,
-                            ]);
-
-                            $questionJson = [
-                                'field' => $child['field'],
-                                'title' => $child['title'],
-                                'guidance' => $child['guidance'],
-                                'required' => $child['required'],
-                            ];
-
-                            $childQuestionVersion = QuestionBankVersion::create([
-                                'question_json' => json_encode($questionJson),
-                                'required' => $child['required'],
-                                'default' => $child['default'],
-                                'question_id' => $childQuestion->id,
-                                'version' => 1,
-                            ]);
-
-                            $questionHasChild = QuestionBankVersionHasChildVersion::create([
-                                'parent_qbv_id' => $questionVersion->id,
-                                'child_qbv_id' => $childQuestionVersion->id,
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Api/V1/QuestionBankController.php
```diff
@@ -411,6 +515,68 @@ public function update(UpdateQuestionBank $request, int $id): JsonResponse
...
             }
 
+            // Don't allow children to also have children, and only allow certain parent types to have children
+            if (!($input['is_child'] ?? false)
+                && isset($input['children'])
+                && in_array($input['field']['component'], ['RadioGroup', 'CheckboxGroup', 'Autocomplete'])) {
+                // Create all children questions and question versions as required.
+                // All must by design have the same version number as the parent - parents and children move versions in lockstep
+                if (isset($input['children'])) {
+                    foreach ($input['children'] as $childListCondition => $childList) {
```
BSnelling: Is this similar enough to the logic in the `store` endpoint to put into a trait and call it in both methods?

database/migrations/2024_11_25_142600_nested_dar_questions.php
```diff
@@ -0,0 +1,50 @@
...
+    public function up(): void
+    {
+        Schema::table('question_bank_questions', function (Blueprint $table) {
+            $table->boolean('is_child')->default(false);
+        });
+        Schema::create('question_bank_version_has_child_version', function (Blueprint $table) {
+            $table->softDeletes();
+            $table->bigInteger('parent_qbv_id')->unsigned();
+            $table->bigInteger('child_qbv_id')->unsigned();
+            $table->string('condition')->nullable();
```
BSnelling, line 20: Is there any specification on what values `condition` can take?  Do you know if the validation of `condition`s will be handled in the BE or FE when an application template is rendered?

spco: My plan is to use string comparison for all conditions. We only support RadioGroup, checkboxGroup, and Autocomplete inputs, so each of them can have their value converted to a string and then we compare with the value in `condition`.

I think FE can handle the condition validation - if you mean checking that a condition is satisfied?

app/Http/Controllers/Api/V1/QuestionBankController.php
spco: Probably yes - I will explore that. I think the only difference will be in the version number handling, which could likely be handled in a Trait with a simple flag.

database/migrations/2024_11_25_142600_nested_dar_questions.php
BSnelling: Yes I was referring to checking if the condition is satisfied.  Makes sense that would be FE.  

And string comparison for conditions makes sense.  As the question bank expands and more people can add to it, I wonder if we will need to worry about monitoring the quality of the conditions (e.g. checking that they match the options in a radioGroup) but for the moment I'd say that's a future concern.

spco: Yes there are a few unknown just yet, so we may require a tweak or two. And I may come to regret string matching but hopefully not!

I anticipate only allowing child questions to be created and updated via their parent question - that's what the controller here does. The only exception is that PATCH doesn't handle nesting, but simply edits any question directly. So we should be able to stipulate that the conditions must match a radiogroup etc value via the FE design - the user has no ability to set the condition independent of its context within the parent.

app/Http/Controllers/Api/V1/QuestionBankController.php
spco: I have pulled out some repeated code to helper functions - would you mind re-reviewing now please? Ta.

database/seeders/QuestionBankSeeder.php
```diff
@@ -19,9 +21,10 @@ public function run(): void
 
 
         DataAccessSection::truncate();
-        QuestionBank::truncate();
-        QuestionBankVersion::truncate();
-
+        QuestionHasTeam::truncate();
+        QuestionBankVersionHasChildVersion::truncate();
+        \DB::table('question_bank_versions')->delete();
+        \DB::table('question_bank_questions')->delete();
```
calmacx, line 27: why delete? foreign key constraints?

BSnelling (approved):

database/seeders/QuestionBankSeeder.php
spco: Yup! This is the best I can do unless I temporarily relax the constraints.

This way feels safer, but it doesn't reset the id counter to 0 - do shout if you think I should do it the other way.

tests/Feature/EmailServiceTest.php
```diff
@@ -21,10 +21,11 @@ class EmailServiceTest extends TestCase
 
     public function setUp(): void
     {
-        parent::setUp([
+        parent::setUp();
```
spco, line 24: This is an unrelated change, but it's a bug that my changes was causing to trigger, so I've fixed it.

JamieByrneHDRUK (approved):

PR #972, opened 2024-12-11
GAT-5830 :: create and update secrets

dnhdruk:
## Screenshots (if relevant)
Screenshot 2024-12-11 at 14 21 41
Screenshot 2024-12-11 at 14 21 51


app/Models/Federation.php
```diff
@@ -17,16 +16,6 @@ class Federation extends Model
...
 
-    protected static function boot()
-    {
-        parent::boot();
-
-        //create a pid for this team
-        static::creating(function ($model) {
-            $model->pid = (string) Str::uuid();
-        });
-    }
```
loki-sinclair-hdruk, line 28: This removal of this, isn't required. In fact, the only part missing was the addition to $fillable on the model. Without that being set, Laravel doesn't know what to do with the extra param. With it set, `boot` will be called first, so this is perfectly valid and then doesn't require updating the controllers above.

dnhdruk: I think `pid` is not necessary because `id` is enough for uniqueness.

dnhdruk: updated based on `id`

loki-sinclair-hdruk: Fair enough, your call. :)

rest.http
```diff
@@ -21,6 +21,7 @@
 # 'developers@hdruk.ac.uk', 'Watch26Task?'
```
loki-sinclair-hdruk: Sorry, missed this. We _really_ need to get rid of these plain text credentials.

dnhdruk: updated

loki-sinclair-hdruk (approved):

BSnelling (approved):

PR #985, opened 2024-12-17
Collections code hardening

BSnelling:
## Describe your changes

Adds new endpoints under v2:
- [GET] /v2/<entity>/?entity_id?
- [GET] /v2/teams/<team_id>/<entity>
- [GET] /v2/teams/1/datasets/status/<archived/draft> 
- [GET] /v2/users/<user_id>/<entity>/?entity_id?
- [GET] /v2/users/<user_id>/<entity>/?entity_id?/status/<archived/draft>
- [GET] /v2/teams/<team_id>/<entity>/?entity_id?
- [GET] /v2/teams/<team_id>/<entity>/?entity_id?/status/<archived/draft>
- [POST/PUT/PATCH/DELETE] /v2/users/<user_id>/<entity>/<entity_id>
- [POST/PUT/PATCH/DELETE] /v2/teams/<team_id>/<entity>/<entity_id>

And rationalises how we implement collection status.  Previously archiving a collection changed the status to `ARCHIVED` and set the `deleted_at` column to a timestamp (`DELETE` effectively acted as `archive`).  In the new endpoints status changes just change the status and the `DELETE` endpoint is only used for actually deleting collections.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-5840


app/Console/Commands/RationaliseStatusDeletedAt.php
```diff
@@ -0,0 +1,39 @@
...
+use Illuminate\Console\Command;
+
+class RationaliseStatusDeletedAt extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'app:rationalise-status-deleted-at';
```
BSnelling, line 16: We will have to run similar commands for other entities (datasets, tools etc).  My intention with this console command is that it can be extended to those other entities in later PRs.

tests/Feature/CollectionTest.php
```diff
@@ -1,900 +0,0 @@
-<?php
```
BSnelling, line 1: This file has been moved to `tests/Feature/V2` and more tests added to it.

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -35,12 +36,271 @@ class CollectionController extends Controller
...
+     */
+    public function index(Request $request): JsonResponse
+    {
+        try {
+            $perPage = $request->has('perPage') ? (int) $request->get('perPage') : Config::get('constants.per_page');
+
+            $sort = $request->query('sort', 'name:desc');
+            $tmp = explode(":", $sort);
+            $sortField = $tmp[0];
+            $sortDirection = array_key_exists(1, $tmp) ? $tmp[1] : 'desc';
```
calmacx: Seeing as you guys have moved to v2, I would suggest implementing/stealing what we have here:
https://github.com/HDRUK/speedi-as-api/blob/dev/app/Traits/SearchManager.php

Then you can consistently call `<Model>::applySorting()`

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -35,12 +36,271 @@ class CollectionController extends Controller
...
+            ->when(
+                $sort,
+                fn ($query) => $query->orderBy($sortField, $sortDirection)
+            )
+            ->paginate((int) $perPage, ['*'], 'page');
+
+            $collections->getCollection()->transform(function ($collection) {
+                if ($collection->image_link && !filter_var($collection->image_link, FILTER_VALIDATE_URL)) {
+                    $collection->image_link = Config::get('services.media.base_url') .  $collection->image_link;
+                }
```
calmacx: consider making a global function you can pass the model and the name of the field that needs prepending

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -35,12 +36,271 @@ class CollectionController extends Controller
...
+
+                return $collection;
+            });
+
+            $collections->getCollection()->transform(function ($collection) {
+                $collection->setAttribute('datasets', $collection->allDatasets  ?? []);
+                return $collection;
+            });
```
calmacx, line 121: Nightmare code... but appreciate you most likely copy & pasted from V1

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -35,12 +36,271 @@ class CollectionController extends Controller
...
+     *    )
+     * )
+     */
+    public function count(Request $request, string $field): JsonResponse
+    {
+        try {
+            $counts = Collection::select($field)
+                ->get()
+                ->groupBy($field)
+                ->map->count();
```
calmacx: Another suggestion to have this as a reusable trait for multiple models

app/Http/Controllers/Api/V2/CollectionController.php
```diff
@@ -35,12 +36,271 @@ class CollectionController extends Controller
...
+     *       ),
+     *    ),
+     * )
+     */
+    public function show(GetCollection $request, int $id): JsonResponse
+    {
+        try {
+            $viewType = $request->query('view_type', 'full');
+            $trimmed = $viewType === 'mini';
+            $withTrashed = false;
```
calmacx: never used? Is it temporary as withTrashed/status is being reworked?

app/Http/Controllers/Api/V2/TeamCollectionController.php
```diff
@@ -0,0 +1,989 @@
...
+     */
+    public function indexDraft(Request $request, int $teamId): JsonResponse
+    {
+        try {
+            $perPage = $request->has('perPage') ? (int) $request->get('perPage') : Config::get('constants.per_page');
+
+            $sort = $request->query('sort', 'name:desc');
+            $tmp = explode(":", $sort);
+            $sortField = $tmp[0];
+            $sortDirection = array_key_exists(1, $tmp) ? $tmp[1] : 'desc';
```
calmacx: example of where code has been copied, and you could use a trait to sort

app/Http/Controllers/Api/V2/TeamCollectionController.php
```diff
@@ -0,0 +1,989 @@
...
+                'image_link',
+                'enabled',
+                'public',
+                'counter',
+                'mongo_object_id',
+                'mongo_id',
+                'team_id',
+                'status',
+            ];
+            $array = $this->checkEditArray($input, $arrayKeys);
```
calmacx, line 824: I never really got this stuff, it could (should?) be handled by EditTeamCollection which is middleware that determines what is accepted in the request to this endpoint

app/Http/Controllers/Api/V2/TeamCollectionController.php
```diff
@@ -0,0 +1,989 @@
...
+
+            if (array_key_exists('keywords', $input)) {
+                $keywords = $input['keywords'];
+                $this->checkKeywords($id, $keywords);
+            }
+
+            // for migration from mongo database
+            if (array_key_exists('created_at', $input)) {
+                Collection::where('id', $id)->update(['created_at' => $input['created_at']]);
+            }
```
calmacx: dont care about this for V2 of the api?

app/Http/Controllers/Api/V2/TeamCollectionController.php
```diff
@@ -0,0 +1,989 @@
...
+            $owningTeamId = $collection->team_id;
+            $this->checkAccess($input, $owningTeamId, null, 'team');
+
+            if ($collection) {
+                CollectionHasDatasetVersion::where(['collection_id' => $id])->delete();
+                CollectionHasTool::where(['collection_id' => $id])->delete();
+                CollectionHasDur::where(['collection_id' => $id])->delete();
+                CollectionHasKeyword::where(['collection_id' => $id])->delete();
+                CollectionHasPublication::where(['collection_id' => $id])->delete();
+                Collection::where(['id' => $id])->delete();
```
calmacx, line 968: Ideally these CollectionsHasX should be deleted upon Collection::delete() cascade so it doesnt have to be done manually (?)

thomgiles (commented):
Looking much better

app/Http/Controllers/Api/V2/CollectionController.php
thomgiles: I agree. That allDatasets stuff needs to go. Agreement with Loki / Emily was it should be datasetHasX etc (I.e. not via versions but for that to happen the flow really needs planning out).

BSnelling: I'm going to leave that as a dedicated task rather than doing it bit by bit.  Hopefully we can incorporate that into the v2 endpoints.

app/Http/Controllers/Api/V2/TeamCollectionController.php
BSnelling: I would have thought the same, but I tried removing it and the `->update($input)` call failed because there were lots of extra array fields in `$input` that the db didn't need.  So I will keep `checkEditArray` in there because it usefully filters out the unneeded array keys.

app/Http/Controllers/Api/V2/TeamCollectionController.php
BSnelling: Yes ideally. That would be a db migration though and I think it will interfere with the behaviour in the v1 endpoints so I'll hold off on that.

app/Http/Controllers/Api/V2/CollectionController.php
BSnelling: This is amazing! I didn't know traits could be used like this.

app/Http/Controllers/Api/V2/TeamCollectionController.php
BSnelling: Good spot, deleted across multiple endpoints.

app/Http/Controllers/Api/V2/TeamCollectionController.php
calmacx: Yeah that seems dodgy - PUT/PATCH/POST shouldnt have stuff in the body that isnt for updating/creating 🤷 

Won't cause you more hassle - but that's a refactor candidate

app/Http/Controllers/Api/V2/CollectionController.php
calmacx: Sounds good

app/Models/Traits/EntityCounter.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace App\Models\Traits;
+
+trait EntityCounter
+{
+    public function scopeApplyCount($query): mixed
```
calmacx, line 7: Im not seeing this used - so just double check

calmacx (approved):

app/Models/Traits/EntityCounter.php
BSnelling: Used in V2/CollectionController.

loki-sinclair-hdruk (approved):

app/Http/Controllers/Api/V2/CollectionController.php
loki-sinclair-hdruk: Further to this, you apply the values iterated over at the model level, and (obvs(?)) include the trait under said model to implement the scoping, per https://github.com/HDRUK/speedi-as-api/blob/dev/app/Models/Organisation.php

dnhdruk (approved):

config/routes_v2.php
```diff
@@ -55,4 +88,186 @@
...
+        'middleware' => [
+            'jwt.verify',
+            'sanitize.input',
+        ],
+        'constraint' => [],
+    ],
+    [
+        'name' => 'collections',
+        'method' => 'get',
+        'path' => '/users/{userId}/collections/status/draft',
```
spco, line 114: See my note in Slack - I think the design of these endpoints needs revisiting

PR #1002, opened 2025-01-07
Start a DAR application with template merging

BSnelling:
## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-5850

### DIFF
--- a/app/Http/Controllers/Api/V1/DataAccessApplicationController.php
+++ b/app/Http/Controllers/Api/V1/DataAccessApplicationController.php
@@ -137,7 +137,7 @@ class DataAccessApplicationController extends Controller
             $input = $request->all();
             $jwtUser = array_key_exists('jwt_user', $input) ? $input['jwt_user'] : [];
 
-            $application = DataAccessApplication::findOrFail($id);
+            $application = DataAccessApplication::where('id', $id)->with('questions')->first();
 
             if ($application) {
                 Auditor::log([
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Api/V1/DataAccessApplicationController.php
```diff
@@ -210,8 +217,82 @@ public function store(CreateDataAccessApplication $request): JsonResponse
...
+                    'dar_application_id' => $application->id
+                ]);
+
+                $gatewayId = $metadata['metadata']['summary']['publisher']['gatewayId'];
+                // check for primary key or pid match...
+                $team = Team::where('id', $gatewayId)->first();
+                if (!$team) {
+                    $team = Team::where('pid', $gatewayId)->first();
+                    if (!$team) {
+                        CloudLogger::write([
```
loki-sinclair-hdruk, line 235: Doesn't Audit determine what method we use for logging?

database/migrations/2025_01_02_121653_create_dar_application_has_dataset_table.php
```diff
@@ -0,0 +1,34 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('dar_application_has_dataset', function (Blueprint $table) {
+            $table->id();
```
loki-sinclair-hdruk: Primary keys don't belong on a pivot, unless you need to manipulate via primary key. In this instance, you're explicitly setting that we have unique constraints on `dar_application_id` and `dataset_id` thus meaning we don't need the primary auto increment. Do you use `dar_application_has_dataset.id` anywhere else?

app/Http/Controllers/Api/V1/DataAccessApplicationController.php
BSnelling: Not sure I understand, are you saying I should use `Auditor` instead of `CloudLogger`? In this scenario I want to throw a warning because the method won't necessarily fail but it likely won't be behaving as expected.  So I figured it would be useful to have a log here that can be seen in GCP for debugging.

database/migrations/2025_01_02_121653_create_dar_application_has_dataset_table.php
BSnelling: The primary key isn't used, I will remove it.

app/Http/Controllers/Api/V1/DataAccessApplicationController.php
loki-sinclair-hdruk: If it's visible in the gcp logs scraped from std::out/err, then it's fine. I think consistently we've always used the Auditor facade, so I was expecting an argument for using CloudLogger instead.

loki-sinclair-hdruk (approved):

dnhdruk (approved):

PR #1007, opened 2025-01-08
fix:(GAT-5945) - encode name if exists for tools,teams

JamieByrneHDRUK:
## Screenshots (if relevant)
image
image

## Describe your changes
So this was originally done here: https://github.com/HDRUK/gateway-api-2/pull/990/files but i missed it off in a few places
## Issue ticket link
https://hdruk.atlassian.net/jira/software/c/projects/GAT/boards/51?assignee=712020%3A05fbbf0f-9386-469a-a212-84cfbf9cc587&selectedIssue=GAT-5945

### DIFF
--- a/app/Http/Controllers/Api/V1/CollectionController.php
+++ b/app/Http/Controllers/Api/V1/CollectionController.php
@@ -720,7 +720,9 @@ class CollectionController extends Controller
                 'status',
             ];
             $array = $this->checkEditArray($input, $arrayKeys);
-
+            if (array_key_exists('name', $input)) {
+                $array['name'] = sanitize_input($input['name']);
+            }
             Collection::where('id', $id)->update($array);
 
             $datasets = array_key_exists('datasets', $input) ? $input['datasets'] : [];
@@ -941,7 +943,9 @@ class CollectionController extends Controller
                     'status',
                 ];
                 $array = $this->checkEditArray($input, $arrayKeys);
-
+                if (array_key_exists('name', $input)) {
+                    $array['name'] = sanitize_input($input['name']);
+                }
                 // Handle the 'deleted_at' field based on 'status'
                 if (isset($input['status']) && ($input['status'] === Collection::STATUS_ARCHIVED)) {
                     $array['deleted_at'] = Carbon::now();
--- a/app/Http/Controllers/Api/V1/CollectionController.php
+++ b/app/Http/Controllers/Api/V1/CollectionController.php
@@ -528,7 +528,7 @@ class CollectionController extends Controller
             $array = $this->checkEditArray($input, $arrayKeys);
 
             if (array_key_exists('name', $input)) {
-                $array['name'] = sanitize_input($input['name']);
+                $array['name'] = format_clean_input($input['name']);
             }
 
             $collection = Collection::create($array);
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Api/V1/CollectionController.php
```diff
@@ -720,7 +720,9 @@ public function update(UpdateTeamCollection $request, int $teamId, int $id): Jso
                 'status',
             ];
             $array = $this->checkEditArray($input, $arrayKeys);
-
+            if (array_key_exists('name', $input)) {
+                $array['name'] = sanitize_input($input['name']);
```
calmacx: Not using middleware, i.e. 
```
'middleware' => [ 'sanitize.input']
```
because only want to sanitise the name? Or this is something extra?

calmacx (commented):
I probably dont have enough context for this one... 

This could cause a mismatch because some data has already been stored without sanitisation and this will just be applied to new create/updates?

JamieByrneHDRUK:
> I probably dont have enough context for this one...
> 
> This could cause a mismatch because some data has already been stored without sanitisation and this will just be applied to new create/updates?

Not really sure I follow what you mean

app/Http/Controllers/Api/V1/CollectionController.php
loki-sinclair-hdruk: Agree, this would be far simpler to just rely on the middleware.

JamieByrneHDRUK: Everything else can have special chars removed, sanitize_input will remove any nasty things like <script>alert('APestyXss!')</script> bit will allow nice things like Macc'ies, we only care about name for that

JamieByrneHDRUK: its using   'sanitize.input', --> [
        'name' => 'collections',
        'method' => 'post',
        'path' => '/teams/{teamId}/collections',
        'methodController' => 'CollectionController@store',
        'namespaceController' => 'App\Http\Controllers\Api\V1',
        'middleware' => [
            'jwt.verify',
            'sanitize.input',
            'check.access:permissions,collections.create',
        ],
        'constraint' => [],
    ]

JamieByrneHDRUK: payload : 
`{name: jamie's}`, 
output for `$input['name']` is` jamie&#39;s`
we want : `jamie's` 
sanitize_input, parses it back but sanitises it

the end points already have 

```
'middleware' => [
'jwt.verify',
'sanitize.input', ------<
'check.access:permissions,collections.create',
]
```
I can rename the helper function sanitize_input to something like format_clean_input? it is confusing


@loki-sinclair-hdruk @calmacx

calmacx: That name makes it less confusing for sure.

I know we've been round the houses on this stuff but I thought we were just 'formatting cleaned inputs' on the FE. So the DB always consistently stores `jamie&39;s` and then wherever this name is used on the FE, the FE needs to decode i.e. `&39;` -> `'` 

Are we saying that this is a pain and we're gonna sanitise inputs and then format these cleaned inputs on the BE only? My 2nd point was that some data has already got into the DB sanitised (but not formatted) - so there could be a mismatch

JamieByrneHDRUK: Yeah, the names are everywhere, spoke to Loki before Christmas and providing its sanitized before going into the db so no xss or nasty sql injects its fine.

Ahh I get you now, so to make it safer if we just do it for those that are created then?

loki-sinclair-hdruk (approved):

calmacx (approved):

PR #1019, opened 2025-01-13
Introduce question_type

BSnelling:
## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-5995

### DIFF
--- a/app/Http/Controllers/Api/V1/QuestionBankController.php
+++ b/app/Http/Controllers/Api/V1/QuestionBankController.php
@@ -259,7 +259,7 @@ class QuestionBankController extends Controller
                 'archived' => $input['archived'] ?? false,
                 'archived_date' => ($input['archived'] ?? false) ? Carbon::now() : null,
                 'is_child' => false,
-                'question_type' => $input['question_type'] ?? 'STANDARD',
+                'question_type' => $input['question_type'] ?? QuestionBank::STANDARD_TYPE,
             ]);
 
             $questionJson = [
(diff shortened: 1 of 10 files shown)


dnhdruk (approved):

app/Http/Controllers/Api/V1/QuestionBankController.php
```diff
@@ -582,7 +592,7 @@ public function edit(EditQuestionBank $request, int $id): JsonResponse
                 ]);
             }
 
-            if (isset($input['team_id'])) {
+            if ($question->question_type === 'CUSTOM') {
```
spco: Previously, all existing questions have a full product of question_id x team_id entries in `QuestionHasTeam` so that all teams could access all questions.

Here (and in the `updateQuestionHasTeams` function) highlights that you're changing the default relationship between Questions and Teams, moving to a model where `QuestionHasTeam` is only filled when the question is custom.

This means we now need a command to scrub the existing `QuestionHasTeam` table contents before we proceed.

app/Http/Controllers/Api/V1/QuestionBankController.php
```diff
@@ -256,6 +259,7 @@ public function store(CreateQuestionBank $request): JsonResponse
                 'archived' => $input['archived'] ?? false,
                 'archived_date' => ($input['archived'] ?? false) ? Carbon::now() : null,
                 'is_child' => false,
+                'question_type' => $input['question_type'] ?? 'STANDARD',
```
spco: Would it be preferable to declare 2 constants, `QuestionBank::TYPE_CUSTOM` and `QuestionBank::TYPE_STANDARD`, in the same way as we do for e.g `Dataset::STATUS_ACTIVE`?

app/Http/Controllers/Api/V1/QuestionBankController.php
BSnelling: Good point, although we haven't made any of this live there could well be data in that table already. I've added a command.

app/Http/Controllers/Api/V1/QuestionBankController.php
BSnelling: Added!

spco (approved):

PR #1023, opened 2025-01-14
GAT-5948 :: Observers for Elastic Indexing

dnhdruk:


README.md
```diff
@@ -23,6 +23,32 @@ If you don't have a mysql client installed you can use the following to deploy a
 
 -   In order to connect to your mysql pod locally (via any other client), you must ensure you port-forward the mysql port `3306` with the following: `kubectl port-forward <deployed-pod-name> 3306:3306`
 
+### Increase the sort_buffer_size to 1G (1 Gigabyte) in your MySQL instance
```
BSnelling, line 26: Does this need to be done because we are adding observers or is it general advice?

app/Observers/CollectionHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,123 @@
...
+            if (!is_null($dataset)) {
+                $this->reindexElastic($dataset->id);
+            }
+        }
+    }
+
+    /**
+     * Handle the CollectionHasDatasetVersion "updated" event.
+     */
+    public function updated(CollectionHasDatasetVersion $collectionHasDatasetVersion): void
```
BSnelling, line 26: This is the same as the `created` method isn't it? Can you write a helper function that both `created` and `updated` can call?

app/Observers/DatasetVersionObserver.php
```diff
@@ -0,0 +1,76 @@
...
+
+        if (!is_null($dataset)) {
+            $this->reindexElastic($dataset->id);
+        }
+    }
+
+    /**
+     * Handle the DatasetVersion "updated" event.
+     */
+    public function updated(DatasetVersion $datasetVersion): void
```
BSnelling, line 24: Here again, can the three functions `created`, `updated` and `deleted` just call a shared helper function?

tests/Feature/DurTest.php
```diff
@@ -277,18 +276,6 @@ public function test_add_new_dur_with_success(): void
 
         $countBefore = Dur::count();
 
-        ECC::shouldReceive("indexDocument")
```
BSnelling, line 280: Can you add some alternative tests that make sure the observer is getting activated and elastic indexing is happening?

README.md
dnhdruk: When I was reindexing to elasticsearch locally, I had a problem with the memory allocated for mysql. I brought here how this could be done locally and it has nothing to do with laravel observers.

app/Observers/DatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/CollectionHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

BSnelling:
@dnhdruk and I were talking about whether Observers will have an impact on performance. I did a very crude test by timing the test suite locally before Observers were added and after. 
- before 63.6s
- after 73.1s
That is not conclusive obviously but I think we should do some careful regression testing in the live env before going ahead with using Observers.  @loki-sinclair-hdruk what do you think?

loki-sinclair-hdruk:
Observers themselves will add a nominal amount of overhead. It's the code they enact which will suffer poor performance. But agree. Some testing before deploy would definitely be wise. Optimisations can no doubt be added.

README.md
loki-sinclair-hdruk: 1GB sort buffer?! Is this for real?? MySql default is 256KB? What could you possibly need a gig of swap for?

spco: Just saying that I regularly have to bump up sort_buffer_size locally from the default. I think it's mostly to do with the BHF dataset that I still have locally (has that been reduced in size in prod?)

loki-sinclair-hdruk: Absolutely, unless you set it within the container, whenever the container restarts it'll lose that config if set any other way. The default is too small, so we set to 256Mb I think. But a Gb, is extreme, so there is clearly some optimisation to be done here as-is.

app/Observers/CollectionHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public function elasticCollectionHasDatasetVersion(CollectionHasDatasetVersion $collectionHasDatasetVersion)
+    {
+        $collectionId = $collectionHasDatasetVersion->collection_id;
+        $collection = Collection::where([
+            'id' => $collectionId,
+            'status' => Collection::STATUS_ACTIVE,
+        ])->first();
+        if (!is_null($collection)) {
```
loki-sinclair-hdruk, line 62: I'd optimise the above call to just `select('id')` which would offer the same `is_null` check and be a fraction of the cost of pulling the entire model, just to check its existence.

app/Observers/CollectionHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+            $this->indexElasticCollections((int) $collectionId);
```
loki-sinclair-hdruk: $collection->id ?

app/Observers/CollectionHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        $datasetVersionId = $collectionHasDatasetVersion->dataset_version_id;
+        $datasetVersion = DatasetVersion::where([
+            'id' => $datasetVersionId
+        ])->first();
+
+        if (!is_null($datasetVersion)) {
```
loki-sinclair-hdruk, line 71: Same deal here

app/Observers/CollectionHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+            $dataset = Dataset::where([
+                'id' => $datasetVersion->dataset_id,
+                'status' => Dataset::STATUS_ACTIVE,
+                ])->first();
+
+            if (!is_null($dataset)) {
```
loki-sinclair-hdruk, line 77: And here

app/Observers/DatasetObserver.php
```diff
@@ -0,0 +1,76 @@
...
+    use IndexElastic;
+
+    /**
+     * Handle the Dataset "created" event.
+     */
+    public function created(Dataset $dataset): void
+    {
+        $datasetVersion = DatasetVersion::where([
+            'dataset_id' => $dataset->id
+        ])->first();
```
loki-sinclair-hdruk: It'll be this that is adding close to 10seconds of lag. You're pulling the entire model again, inclusive of metadata, and doing nothing with it. Just `select('id')` and check `is_null` on that instead and this will fly. You're only passing `$dataset->id` to `reindexElastic` and thus this entire model for a null check is wildly expensive.

app/Observers/DatasetObserver.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Handle the Dataset "updated" event.
+     */
+    public function updated(Dataset $dataset): void
+    {
+        $prevStatus = $dataset->prevStatus;
+        $datasetVersion = DatasetVersion::where([
+            'dataset_id' => $dataset->id
+        ])->first();
```
loki-sinclair-hdruk: Same here

app/Observers/DurHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+        //
+    }
+
+    public function elasticDurHasDatasetVersion(DurHasDatasetVersion $durHasDatasetVersion)
+    {
+        $durId = $durHasDatasetVersion->dur_id;
+        $dur = Dur::where([
+            'id' => $durId,
+            'status' => Dur::STATUS_ACTIVE,
+        ])->first();
```
loki-sinclair-hdruk: Same here

app/Observers/DurHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+        if (!is_null($dur)) {
+            $this->indexElasticDur($dur->id);
+        }
+
+        $datasetVersionId = $durHasDatasetVersion->dataset_version_id;
+        $datasetVersion = DatasetVersion::where([
+            'id' => $datasetVersionId
+        ])->first();
```
loki-sinclair-hdruk: And here

app/Observers/DurHasDatasetVersionObserver.php
```diff
@@ -0,0 +1,82 @@
...
+
+        if (!is_null($datasetVersion)) {
+            $dataset = Dataset::where([
+                'id' => $datasetVersion->dataset_id,
+                'status' => Dataset::STATUS_ACTIVE,
+            ])->first();
```
loki-sinclair-hdruk: Going to stop commenting now, as I'd imagine all of these need updating to be leaner. Give me a shout if you don't get it.

loki-sinclair-hdruk:
@BSnelling I've done a review, the main problem and artificial degradation is due to pulling back entire models to run `is_null` checks on them, when the same can be done on a `select('id')` for each, thus reducing footprint and wasted cycles.

@dnhdruk Have added a number of comments, and rather than continue, suggest you update any db calls pulling back the entire model, just just the ID where applicable and null check that instead. These observers will fly after that. No need to pull multiples of DatasetVersion when all we want to do is check its existence.

app/Observers/CollectionHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/CollectionHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/CollectionHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/CollectionHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/DatasetObserver.php
dnhdruk: an update has been sent.

app/Observers/DatasetObserver.php
dnhdruk: an update has been sent.

app/Observers/DurHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/DurHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

app/Observers/DurHasDatasetVersionObserver.php
dnhdruk: an update has been sent.

README.md
loki-sinclair-hdruk: Is this still required, even now we don't pull entire models? Have you tried? It seems excessive still.

app/Models/Collection.php
```diff
@@ -31,6 +34,8 @@ class Collection extends Model
     public const STATUS_DRAFT = 'DRAFT';
     public const STATUS_ARCHIVED = 'ARCHIVED';
 
+    public $prevStatus;
```
loki-sinclair-hdruk: no type

app/Models/Dataset.php
```diff
@@ -31,6 +34,8 @@ class Dataset extends Model
     public const ORIGIN_API = 'API';
     public const ORIGIN_GMI = 'GMI';
 
+    public $prevStatus;
```
loki-sinclair-hdruk: and here

app/Models/Dur.php
```diff
@@ -23,6 +26,8 @@ class Dur extends Model
     public const STATUS_DRAFT = 'DRAFT';
     public const STATUS_ARCHIVED = 'ARCHIVED';
 
+    public $prevStatus;
```
loki-sinclair-hdruk: and here

app/Models/Publication.php
```diff
@@ -3,24 +3,28 @@
...
     use SoftDeletes;
     use Prunable;
     use DatasetFetch;
 
-
     public const STATUS_ACTIVE = 'ACTIVE';
     public const STATUS_DRAFT = 'DRAFT';
     public const STATUS_ARCHIVED = 'ARCHIVED';
 
+    public $prevStatus;
```
loki-sinclair-hdruk: and here

app/Models/Collection.php
dnhdruk: the code was updated

app/Models/Dataset.php
dnhdruk: the code was updated

app/Models/Dur.php
dnhdruk: the code was updated

dnhdruk: the code was updated

app/Models/Publication.php
dnhdruk: the code was updated

loki-sinclair-hdruk (approved):

BSnelling (approved):

PR #1039, opened 2025-01-23
Convert more QB endpoints to new syntax

spco:
## Describe your changes
- More endpoints now use the new syntax, and support `is_child` and `section_id` filtering, and pagination.
- Moved `getVersion` to a Trait for reuse across controllers


app/Http/Controllers/Api/V1/QuestionBankController.php
```diff
@@ -34,11 +36,55 @@ class QuestionBankController extends Controller
      *      tags={"QuestionBank"},
      *      summary="QuestionBank@index",
      *      security={{"bearerAuth":{}}},
+     *      @OA\Parameter(
+     *          name="section_id",
```
BSnelling, line 40: What is the FE use case for being able to filter all categories of question (standard, custom, archived) by section? 

There is an endpoint `/teams/{teamId}/questions/section/{sectionId}` which returns a team's questions from a specific section.  Should we consider an endpoint that follows the same design pattern, listing all questions from a specified section?

spco: TBF, I don't think we need that specific filter as it stands. Filtering by section is needed in the template builder, which will use the `/teams/{teamId}/questions/section/{sectionId}` endpoint. 

For MVP, we are relying on hdr users. These aren't part of all teams and thus won't have a relevant entry in `QuestionHasTeam`, to manage custom questions etc. So I added this filter to this endpoint so we can use a team-agnostic endpoint for them to list questions of each type, but tbf I don't think this will be used. Want me to remove it for clarity?

Post-MVP things might be a bit different - we need to have a conversation of what that looks like, particularly how hdr users might use roles for permissions.

BSnelling: No we can keep it in if it might be useful. I'm just wary of the `index` endpoints becoming like the one in datasets, which had lots of query params and filtering, and which was refactored as part of code hardening.

BSnelling (approved):

dnhdruk (approved):

PR #1056, opened 2025-02-03
Bug: GAT-6251 publication-not-appearing

thomgiles:
## Describe your changes

Added description field to publication table to distinguish between publications added by the LinkageExtractor job and the publication controller. 

Added Publication Handling to the LinkageExtractor job. 

Fixed Logic in the publication controller leading to the bug switching relationship type. 

Functionality tested and working in DEV. 

FYI I hate softDeletes() 🗡️ 

## Screenshots (if relevant)
Screenshot 2025-02-07 at 10 29 03 AM

Screenshot 2025-02-07 at 10 30 15 AM

Yes. One small table change

- [x] I have added Swagger annotations for new endpoints (if applicable)


database/migrations/2025_02_03_115734_add_description_to_publication_has_dataset_version_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+class AddDescriptionToPublicationHasDatasetVersionTable extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('publication_has_dataset_version', function (Blueprint $table) {
+            $table->text('description')->nullable()->after('link_type');
```
dnhdruk, line 15: please update requests: CreatePublication, UpdatePublication, EditPublication

app/Http/Controllers/Api/V1/PublicationController.php
```diff
@@ -910,17 +910,15 @@ private function getPublicationById(int $publicationId)
     private function checkDatasets(int $publicationId, array $inDatasets)
```
dnhdruk, line 913: if you are going to add the same dataset id after a new version, I think you should check and update the link with the new version versus the old version

database/migrations/2025_02_03_115734_add_description_to_publication_has_dataset_version_table.php
thomgiles: I can but there is really no need. its always null when triggered from the pub controller. description is only set if triggered via the Job. this is the same as for DvHasDv.

app/Http/Controllers/Api/V1/PublicationController.php
thomgiles: You mean expand the delete so that it covers all other dataset versions. So relations are only re-set to the current version?

I think that goes beyond the scope of this Bug. I.e. I think versioning behaviour still needs planning out. I mean, really when DV changes there should probably be a separate background job that runs to update all the dependences from version x to version y.

database/migrations/2025_02_03_115734_add_description_to_publication_has_dataset_version_table.php
thomgiles: Added in on latest commit

app/Jobs/LinkageExtraction.php
```diff
@@ -9,128 +9,226 @@
...
-                    'direct_linkage' => 1,
-                    'description' => $this->description
-                ]);
-
+                foreach ($data as $d) {
+                    $targetDatasetVersionId = $this->findTargetDataset($d);
+                    if (!$targetDatasetVersionId) {
+                        continue;
+                    }
+                    DatasetVersionHasDatasetVersion::updateOrCreate([
```
spco: `updateOrCreate` needs 2 parameters, otherwise it is a no-op. Here, you try to find an entry, and then you do nothing to it.

app/Http/Controllers/Api/V1/PublicationController.php
```diff
@@ -910,17 +913,15 @@ private function getPublicationById(int $publicationId)
...
 
-        $pubs = PublicationHasDatasetVersion::where(['publication_id' => $publicationId])->get();
-        foreach ($pubs as $pub) {
-            $dataset_id = DatasetVersion::where("id", $pub->dataset_version_id)->first()->dataset_id;
-            if (!in_array($dataset_id, $this->extractInputIdToArray($inDatasets))) {
-                $this->deletePublicationHasDatasetVersions($publicationId, $pub->dataset_version_id);
-            }
-        }
+        // There was an error here where by the Publications were not getting cleared out / exhibiting
+        // wierd cache like behavior on the FE. Clearing all links for a Pub and restoring active ones
```
spco:
```suggestion
        // weird cache like behavior on the FE. Clearing all links for a Pub and restoring active ones
```

spco (changes requested):

app/Jobs/LinkageExtraction.php
thomgiles: I did not write that section of the code. @calmacx did. ;) It could be switched out for DatasetVersionHasDatasetVersion::firstOrCreate() which is probably better as all links from that version are deleted before re-linkage...

spco (approved):

dnhdruk (approved):

PR #1074, opened 2025-02-11
fix(GAT-6386): Getting GMI working

calmacx:
## Describe your changes

- Separate methods for metadatafed
- Cater for data coming in from metadata fed
- use traser data for ted
- when deleting and updating use team id as well

## Issue ticket link
https://hdruk.atlassian.net/jira/software/c/projects/GAT/boards/51?assignee=712020%3A05fbbf0f-9386-469a-a212-84cfbf9cc587&selectedIssue=GAT-6386


app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -1011,7 +1010,7 @@ public function edit(EditDataset $request, int $id)
      *      )
      * )
      */
-    public function destroy(DeleteDataset $request, string $id) // softdelete
+    public function destroy(Request $request, string $id) // softdelete
```
calmacx, line 1019:
```

class DeleteDataset extends BaseFormRequest
{
    /**
     * Get the validation rules that apply to the request.
     *
     * @return array<string, \Illuminate\Contracts\Validation\ValidationRule|array<mixed>|string>
     */
    public function rules(): array
    {
        return [
            'id' => [
                'required',
                'int',
                'exists:datasets,id',
            ],
        ];
    }
```

what's the point in that? when the function is called with ` destroy(Request $request, string $id)`

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -834,13 +834,15 @@ public function destroy(DeleteDataset $request, int $id) // softdelete
     public function destroyByPid(Request $request, string $pid) // softdelete
     {
         $dataset = Dataset::where('pid', "=", $pid)->first();
-        return $this->destroy($request, $dataset->id);
+        $updatedRequest = $request->merge(['team_id' => $dataset->team_id]);
+        return $this->destroy($updatedRequest, $dataset->id);
```
calmacx: this new API needs the `team_id` now - but we dont know that when we're using the service layer to updated/dete datasets by PID via Federation

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -834,13 +834,15 @@ public function destroy(DeleteDataset $request, int $id) // softdelete
...
     }
 
-    public function updateByPid(UpdateDataset $request, string $pid)
```
calmacx: `UpdateDataset` requires us to pass `id` but we dont know it!! we only know `pid` !

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -834,13 +834,15 @@ public function destroy(DeleteDataset $request, int $id) // softdelete
...
+    public function updateByPid(Request $request, string $pid)
     {
         $dataset = Dataset::where('pid', "=", $pid)->first();
-        return $this->update($request, $dataset->id);
+        $updatedRequest = $request->merge(['team_id' => $dataset->team_id]);
+        return $this->update($updatedRequest, $dataset->id);
```
calmacx: As above - may need some thinking... we dont know the team_id ... unless we start supplying it in the payload from the GMI service

Need to make sure the permissions checks are ok here - so that only the superuser or the service user can use these endpoints.. may be better to add that as a JWT middleware - only admins can use this endpoint

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -730,8 +729,8 @@ public function update(UpdateDataset $request, int $id)
 
             $inputSchema = isset($input['metadata']['schemaModel']) ? $input['metadata']['schemaModel'] : null;
             $inputVersion = isset($input['metadata']['schemaVersion']) ? $input['metadata']['schemaVersion'] : null;
-
-            $submittedMetadata = $input['metadata']['metadata'];
+            //TODO go over with big c
+            $submittedMetadata = ($input['metadata']['metadata'] ?? $input['metadata']);
```
JamieByrneHDRUK, line 733: This was null pointing without this locally for me. there was no second ['metadata'] @calmacx  is this down to my local data ?

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -766,6 +765,10 @@ public function update(UpdateDataset $request, int $id)
             ]);
 
             $versionNumber = $currDataset->lastMetadataVersionNumber()->version;
+            //TODO go over with big c
+            if (!is_array($submittedMetadata)) {
```
JamieByrneHDRUK, line 770: Same here this was not an array just a JSON string, its used as the third param for updateMetadataVersion which requires it be an array, again is this due to my local data @calmacx  ?

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -1053,6 +1056,38 @@ public function destroyByPid(Request $request, string $pid) // softdelete
         return $this->destroy($request, $dataset->id);
     }
 
+    public function federationDestroyByPid(Request $request, string $pid) // softdelete
+    {
+        $input = $request->all();
+        $teamId = (int)$input['team_id'];
+        $this->checkAccess($input, $teamId, null, 'team');
+        $dataset = Dataset::where('pid', '=', $pid)
+                  ->where('team_id', '=', $teamId)
```
JamieByrneHDRUK, line 1068: Note the change here to also include the teamID, I was erroring locally as the PID existed for multiple teams

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -1053,6 +1056,38 @@ public function destroyByPid(Request $request, string $pid) // softdelete
...
+
+        return $this->destroy($request, $dataset->id);
+    }
+
+    public function federationUpdateByPid(UpdateDataset $request, string $pid)
+    {
+        $input = $request->all();
+        $teamId = (int) $input['team_id'];
+        $dataset = Dataset::where('pid', '=', $pid)
+                  ->where('team_id', '=', $teamId)
```
JamieByrneHDRUK, line 1083: Note the change here to also include the teamID, I was erroring locally as the PID existed for multiple teams

app/Http/Controllers/Api/V1/DatasetController.php
JamieByrneHDRUK: This also happened in dev when i reverted this line

app/Http/Controllers/Api/V1/DatasetController.php
JamieByrneHDRUK: This also happened in dev when i reverted this line

app/Http/Controllers/Api/V1/DatasetController.php
JamieByrneHDRUK: I think when coming in from fed its not nested

app/Http/Controllers/Api/V1/DatasetController.php
JamieByrneHDRUK: I think this is due to the payload in metadata fed

BSnelling (approved):

app/Http/Controllers/Api/V1/DatasetController.php
```diff
@@ -731,9 +730,9 @@ public function update(UpdateDataset $request, int $id)
             $inputSchema = isset($input['metadata']['schemaModel']) ? $input['metadata']['schemaModel'] : null;
             $inputVersion = isset($input['metadata']['schemaVersion']) ? $input['metadata']['schemaVersion'] : null;
 
-            $submittedMetadata = $input['metadata']['metadata'];
+            $submittedMetadata = ($input['metadata']['metadata'] ?? $input['metadata']);
             $gwdmMetadata = null;
-
+            $useGWDMetada = false;
```
calmacx, line 735: typo on variable name - not very important

JamieByrneHDRUK: Absolute disaster

JamieByrneHDRUK (approved):

JamieByrneHDRUK:
Big C Approved

PR #1151, opened 2025-03-24
fix!(GAT-1234): convert metadatain json format

dnhdruk:
## Screenshots (if relevant)
Screenshot 2025-03-24 at 10 33 57
Screenshot 2025-03-24 at 10 35 28

run
```
php artisan app:update-metadata-to-same-format
```

yes

### DIFF
--- a/app/Console/Commands/UpdateMetadataToSameFormat.php
+++ b/app/Console/Commands/UpdateMetadataToSameFormat.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\DatasetVersion;
+use Illuminate\Console\Command;
+
+class UpdateMetadataToSameFormat extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'app:update-metadata-to-same-format';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Command description';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $datasetVersions = DatasetVersion::select('id')->get();
+
+        foreach ($datasetVersions as $datasetVersion) {
+            $metadata = \DB::table('dataset_versions')
+                ->where('id', $datasetVersion->id)
+                ->select('id', 'dataset_id', 'metadata', \DB::raw('JSON_TYPE(metadata) as metadata_type'))
+                ->first();
+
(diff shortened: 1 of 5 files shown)


app/Console/Commands/UpdateMetadataToSameFormat.php
```diff
@@ -0,0 +1,50 @@
...
+        $datasetVersions = DatasetVersion::select('id')->get();
+
+        foreach ($datasetVersions as $datasetVersion) {
+            $metadata = \DB::table('dataset_versions')
+                ->where('id', $datasetVersion->id)
+                ->select('id', 'dataset_id', 'metadata', \DB::raw('JSON_TYPE(metadata) as metadata_type'))
+                ->first();
+
+            if ($metadata->metadata_type === 'STRING') {
+                $arrayMetadata = json_decode(json_decode($metadata->metadata), true);
```
JamieByrneHDRUK, line 38: do you need to decode twice?

BSnelling, line 38: Is the double `json_decode` always needed? It won't fail if it is not needed will it?

app/Models/DatasetVersion.php
```diff
@@ -43,6 +44,25 @@ class DatasetVersion extends Model
         'version',
     ];
 
+    /**
+     * Get and Set the metadata.
+     *
+     * @return \Illuminate\Database\Eloquent\Casts\Attribute
+     */
+    public function metadata(): Attribute
```
BSnelling, line 52: Does this replace the `getMetadataAttribute` function completely?  So `$dataset->metadata` will still return an array?

app/Console/Commands/UpdateMetadataToSameFormat.php
dnhdruk: string to json -> json to array

dnhdruk: it is double encoded.
in db is a string

app/Models/DatasetVersion.php
dnhdruk: when we get data from db return it is an array converted with `get` from `DatasetVersion` model and method `metadata()`

dnhdruk: you can use test in local

BSnelling: LGTM from a few local tests.

BSnelling (approved):

JamieByrneHDRUK (approved):

PR #1158, opened 2025-03-25
feat(GAT-6577): Add submission stats to endpoints and allow nulling approval_status

BSnelling:
## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-6577

### DIFF
--- a/app/Http/Requests/DataAccessApplication/EditDataAccessApplication.php
+++ b/app/Http/Requests/DataAccessApplication/EditDataAccessApplication.php
@@ -37,6 +37,7 @@ class EditDataAccessApplication extends BaseFormRequest
             ],
             'approval_status' => [
                 'string',
+                'nullable',
                 'in:APPROVED,APPROVED_COMMENTS,FEEDBACK,REJECTED',
             ],
         ];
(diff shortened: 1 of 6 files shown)


dnhdruk (approved):

app/Http/Requests/DataAccessApplication/EditDataAccessApplication.php
```diff
@@ -37,6 +37,7 @@ public function rules(): array
             ],
             'approval_status' => [
                 'string',
+                'nullable',
```
loki-sinclair-hdruk, line 40: Is this allowed when its supposed to be `in:APPROVED, APPROVED_COMMENTS,FEEDBACK,REJECTED`? Would `sometimes` not be a better fit? Then it's not required, and the validation runs if supplied.

loki-sinclair-hdruk (approved):

app/Http/Requests/DataAccessApplication/EditDataAccessApplication.php
BSnelling: This is for the scenario where the custodian explicitly pushes a submitted application back to "DRAFT" and sets the approval status to null, e.g. PATCH with the payload
```
{
    "submission_status": "DRAFT",
    "approval_status": null
}
```
Would a `sometimes` rule work in that scenario?

BSnelling: I tried it out with `sometimes` instead of `nullable` in the rule but I think that doesn't work for the scenario where we want to change the `approval_status` from e.g. `APPROVED` back to `null`.  Happy to revisit if I've misunderstood.

PR #1159, opened 2025-03-25
feat(GAT-6612): endpoint for user to delete a DAR

BSnelling:
## Describe your changes

...and some extra info in the dar reviews endpoints

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-6612

### DIFF
--- a/app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
+++ b/app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
@@ -1208,8 +1208,8 @@ class UserDataAccessApplicationController extends Controller
         $jwtUser = array_key_exists('jwt_user', $input) ? $input['jwt_user'] : [];
 
         try {
-            $application = DataAccessApplication::where('id', $id)->with('teams')->first();
-            if (($jwtUser['id'] != $userId) || ($jwtUser['id'] != $application->applicant_id)) {
+            $application = DataAccessApplication::with('teams')->findOrFail($id);
+            if (($jwtUser['id'] !== $userId) || ($jwtUser['id'] !== $application->applicant_id)) {
                 throw new UnauthorizedException('User does not have permission to use this endpoint to delete this file.');
             }
 
--- a/app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
+++ b/app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
@@ -12,6 +12,7 @@ use App\Exceptions\UnauthorizedException;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\DataAccessApplication\CreateUserDataAccessApplicationAnswer;
 use App\Http\Requests\DataAccessApplication\EditUserDataAccessApplication;
+use App\Http\Requests\DataAccessApplication\DeleteUserDataAccessApplication;
 use App\Http\Requests\DataAccessApplication\DeleteUserDataAccessApplicationFile;
 use App\Http\Requests\DataAccessApplication\GetUserDataAccessApplication;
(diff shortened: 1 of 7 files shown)


dnhdruk (approved):

app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
```diff
@@ -1147,6 +1149,107 @@ public function destroyFile(DeleteUserDataAccessApplicationFile $request, int $u
...
+     *      )
+     * )
+     */
+    public function destroy(DeleteUserDataAccessApplication $request, int $userId, int $id): JsonResponse
+    {
+        $input = $request->all();
+        $jwtUser = array_key_exists('jwt_user', $input) ? $input['jwt_user'] : [];
+
+        try {
+            $application = DataAccessApplication::where('id', $id)->with('teams')->first();
```
calmacx: `findOrFail($id)` would be safer

app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
```diff
@@ -1147,6 +1149,107 @@ public function destroyFile(DeleteUserDataAccessApplicationFile $request, int $u
...
+            if (($jwtUser['id'] != $userId) || ($jwtUser['id'] != $application->applicant_id)) {
```
calmacx: `!==`

app/Http/Traits/DataAccessApplicationHelpers.php
```diff
@@ -387,4 +387,22 @@ public function getDarManagers(int $teamId): ?array
 
         return $users;
     }
+
+    public function getNamesInComments(Collection $reviews): void
+    {
+        foreach ($reviews as $review) {
+            foreach ($review['comments'] as $comment) {
+                if ($comment['team_id']) {
+                    $comment['team_name'] = Team::where('id', $comment['team_id'])->value('name');
```
calmacx: doesnt seem efficient to be making an SQL call within a loop within a loop - especially if there are lots of reviews and lots of comments (not sure how likely that is..)

app/Http/Traits/DataAccessApplicationHelpers.php
```diff
@@ -387,4 +387,22 @@ public function getDarManagers(int $teamId): ?array
...
+                    $comment['team_name'] = null;
+                }
+                if ($comment['user_id']) {
+                    $comment['user_name'] = User::where('id', $comment['user_id'])->value('name');
+                } else {
+                    $comment['user_name'] = null;
+                }
+            }
+        }
+    }
```
calmacx: think there would be a way to loop all review_comments, pull out unique `team_id` and `user_id` and then make a single call to `Team` and `User` with a `whereIn` - then do a map lookup

calmacx (changes requested):
Minor changes - plus consider optimising SQL calls buried in for-loops

app/Http/Controllers/Api/V1/UserDataAccessApplicationController.php
dnhdruk: it is not necessary because the validation is already defined in form request.

app/Http/Traits/DataAccessApplicationHelpers.php
dnhdruk: Cyclomatic complexity seems to be high here but I cannot find this method somewhere used.

dnhdruk: maybe you can replace entire method with an observer.

BSnelling: Updated the code to use attributes on the `Comment` model instead. This should be more efficient than the SQL in loops.

calmacx (approved):

PR #1218, opened 2025-05-07
feat(GAT-6927): feature flagging with enquries

JamieByrneHDRUK:


app/Http/Controllers/Api/V1/EnquiryThreadController.php
```diff
@@ -240,9 +242,51 @@ public function store(Request $request): JsonResponse
...
+                    }
+                }
+            } else {
+
+                foreach ($payload['thread']['datasets'] as $d) {
+                    $t = Team::where('id', $d['team_id'])->first();
+                    $dataCustodians[] = $t->name;
+                }
             }
             $payload['thread']['dataCustodians'] = array_unique($dataCustodians);
```
BSnelling, line 298: I think lines 295-301 also need to go inside the `else`.  Otherwise the `teamIds` and `teamNames` from the `(Feature::active('features.SDEConciergeServiceEnquiry')) === true` case will be overwritten.

tests/Feature/EnquiryThreadTest.php
```diff
@@ -331,4 +341,244 @@ public function test_add_new_dar_enquiry_thread_with_success(): void
...
+        ]);
+
+        $numThreadsAfter = EnquiryThread::count();
+
+        // Test that an email was sent to the network concierge
+        Queue::assertPushed(function (SendEmailJob $email) {
+            return $email->to['to']['name'] === 'SDE Network Concierge';
+        });
+
+        $this->assertEquals($numThreadsAfter, $numThreadsBefore + count($datasets));
```
JamieByrneHDRUK, line 481: @BSnelling this is what i changed from +1 to the count of the datasets. I think/hope this is correct

BSnelling: There was an update brought in after the Concierge stuff was initially implemented and reverted, which spawned separate enquiry threads for each team in an enquiry. So using the count of datasets is correct here.

BSnelling (approved):

dnhdruk (approved):

tests/Feature/EnquiryThreadTest.php
```diff
@@ -331,4 +341,244 @@ public function test_add_new_dar_enquiry_thread_with_success(): void
...
+
+        $numThreadsAfter = EnquiryThread::count();
+
+        // Test that an email was NOT sent to the network concierge
+        Queue::assertPushed(function (SendEmailJob $email) {
+            return $email->to['to']['name'] !== 'SDE Network Concierge';
+        });
+    }
+
+    private function createUser(): int
```
loki-sinclair-hdruk, line 523: Can you make a note/ticket to move these to TestCase.php a level up, as we've got a few of these now and it would be better to make them reusable under `$this->createUser()`, rather than reinvent the wheel. Not for now, obvs, just a note to ensure it's looked at soonish to avoid more duplication.

loki-sinclair-hdruk (approved):

tests/Feature/EnquiryThreadTest.php
JamieByrneHDRUK: Will do

PR #1229, opened 2025-05-12
feat(GAT-6927): feature flag endpoint for action

JamieByrneHDRUK:
## Screenshots (if relevant)

image
image
image


app/Http/Controllers/Api/V1/FeatureFlagController.php
```diff
@@ -0,0 +1,106 @@
...
+{
+    /**
+     * @OA\Post(
+     *    path="/api/v1/feature-flags",
+     *    operationId="define_feature_flags_from_github",
+     *    tags={"Application"},
+     *    summary="Fetch and define feature flags from GitHub",
+     *    description="Validates a GitHub webhook signature, fetches a JSON config from the configured GitHub URL, and defines feature flags using Laravel Pennant.",
+     *    @OA\Parameter(
+     *        name="X-Hub-Signature-256",
```
JamieByrneHDRUK: bin

app/Http/Controllers/Api/V1/FeatureFlagController.php
```diff
@@ -0,0 +1,155 @@
...
+    public function index(Request $request, FeatureFlagManager $flagManager): JsonResponse
+    {
+        $featureFlagToken = env('FEATURE_FLAG_API_TOKEN');
+        $authHeader = $request->header('Authorization');
+
+        if (!$authHeader || !str_starts_with($authHeader, 'Bearer ')) {
+            return response()->json(['message' => 'Unauthorized: Missing or invalid Authorization header.'], 401);
+        }
+
+        $providedToken = substr($authHeader, 7); // remove "Bearer "
```
dnhdruk, line 63: Just split the string based on space.
check if is bearer first string and take the second part

dnhdruk (approved):

app/Http/Controllers/Api/V1/EnquiryThreadController.php
```diff
@@ -245,7 +245,7 @@ public function store(Request $request): JsonResponse
             $teamIds = [];
             $teamNames = [];
 
-            if (Feature::active('features.SDEConciergeServiceEnquiry')) {
+            if (Feature::active('SDEConciergeServiceEnquiry')) {
```
dnhdruk, line 248: Maybe you need a new method here, and replace/move the code in the new method

app/Http/Controllers/Api/V1/FeatureFlagController.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        $featureFlags = $request->json()->all();
+
+        if (!empty($featureFlags)) {
+            Log::info("Using feature flags from request body: " . print_r($featureFlags, true));
+        } else {
+            $url = env('FEATURE_FLAGGING_CONFIG_URL');
+
+            if (app()->environment('testing') || !$url) {
```
BSnelling: Should this check go outside the `if (!empty($featureFlags))` statement?  I'm wondering if there is a scenario where we have a test that calls the endpoint with a body.

app/Http/Controllers/Api/V1/EnquiryThreadController.php
JamieByrneHDRUK: I've changed the json so it no longer defines it with features. just the raw key. I thought it looks better this way

app/Http/Controllers/Api/V1/FeatureFlagController.php
```diff
@@ -0,0 +1,155 @@
...
+         *    @OA\Response(
+         *        response=200,
+         *        description="List of enabled feature flags",
+         *        @OA\JsonContent(
+         *            @OA\Property(property="enabled_features", type="array", @OA\Items(type="string"))
+         *        )
+         *    )
+         * )
+         */
+    public function getEnabledFeatures(Request $request, FeatureFlagManager $flagManager): JsonResponse
```
BSnelling, line 115: Might we want to be able to list the status of all features (enabled or disabled)?

config/routes.php
```diff
@@ -390,6 +390,24 @@
...
+        'middleware' => [],
+        'constraint' => [],
+    ],
+    [
+        'name' => 'flags',
+        'method' => 'get',
+        'path' => '/feature-flags/enabled',
+        'methodController' => 'FeatureFlagController@getEnabledFeatures',
+        'namespaceController' => 'App\Http\Controllers\Api\V1',
+        'middleware' => [],
```
BSnelling, line 408: Do we want to add any auth middleware to this endpoint? JWT for instance.

app/Services/FeatureFlagManager.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Laravel\Pennant\Feature;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\Log;
+
+class FeatureFlagManager
+{
+    public const CACHE_KEY = 'defined_feature_flags';
+
+    public function define(array $flags, string $prefix = ''): void
```
BSnelling, line 12: What is the use case for the `prefix`?  Are we planning to have different categories of features in the future?

config/routes.php
JamieByrneHDRUK: I wouldn't recommend adding auth layer in the call from the frontend would be server side

app/Services/FeatureFlagManager.php
JamieByrneHDRUK: Good spot this would be for nested features like:
```
 "gmi": {
      "enabled": true,
      "features": {
        "auth": {
          "enabled": true
        },
        "no-auth": {
          "enabled": false
        },
        "privateApps": {
          "enabled": false
        }
      }
    }
```

BSnelling: Makes sense, thanks!

app/Http/Controllers/Api/V1/FeatureFlagController.php
JamieByrneHDRUK: Body has now gone

BSnelling (approved):

PR #1234, opened 2025-05-12
fix(GAT-6993): For researcher add related applications to list

BSnelling:
## Screenshots (if relevant)

dar_joint_apps2
dar_joint_apps1

## Describe your changes

Updated so that when a joint application is submitted, the applications related to each other are still displayed in groups, even when their statuses are different.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-6993

None

No

### DIFF
--- a/app/Http/Traits/DataAccessApplicationHelpers.php
+++ b/app/Http/Traits/DataAccessApplicationHelpers.php
@@ -424,10 +424,10 @@ trait DataAccessApplicationHelpers
         $applications->getCollection()->transform(function ($application) {
             $sameProject = DataAccessApplication::where('project_id', $application['project_id'])
                 ->get();
-            $teams = array();
-            foreach ($sameProject as $app) {
-                $teams = array_merge($teams, $app['teams']->toArray());
-            }
+            $teams = $sameProject->flatMap(function ($app) {
+                return $app->teams;
+            })->unique('id')->values();
+
             $application['teams'] = $teams;
             return $application;
         });
--- a/app/Http/Traits/DataAccessApplicationHelpers.php
+++ b/app/Http/Traits/DataAccessApplicationHelpers.php
@@ -421,10 +421,7 @@ trait DataAccessApplicationHelpers
 
     public function returnApplicationsInProject(LengthAwarePaginator $applications): LengthAwarePaginator
     {
-        $applicationsResult = array();
-
-        foreach ($applications as $application) {
-            $applicationArray = array();
+        $applications->getCollection()->transform(function ($application) {
             $sameProject = DataAccessApplication::where('project_id', $application['project_id'])
                 ->get();
             $teams = array();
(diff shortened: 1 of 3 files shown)


dnhdruk (approved):

app/Http/Traits/DataAccessApplicationHelpers.php
```diff
@@ -418,4 +418,31 @@ public function groupApplicationsByProject(LengthAwarePaginator $applications):
...
+
+        $page = $applications::resolveCurrentPage();
+        $perPage = Config::get('constants.per_page');
+
+        return new LengthAwarePaginator(
+            $applicationsResult,
+            count($applicationsResult),
+            $perPage,
+            $page
+        );
```
calmacx: Looks like `applicationsResult` will be always the same length as `applications` so dont need to reconstruct the LengthAwarePaginatior?

So you could transform instead of reconstruct:
```
    $applications->getCollection()->transform(function ($application) {
        ...
       <modify teams>
       ..
       }
      return $applications
```


Making SQL calls within the forloop isnt very optimal either, maybe pull that outside the loop:
```
$projectIds = $applications->getCollection()->pluck('project_id')->unique();
```
then can load them all via `->whereIn('project_id',$projectIds)` to lookup. I get that you might not have many applications per project though... so might be overkill.

calmacx (changes requested):

app/Http/Traits/DataAccessApplicationHelpers.php
BSnelling: Good shout on using `transform`, that's much nicer!  

I don't think it is worth pulling the lookup out of the loop in this case. We're not expecting many applications per project.

app/Http/Traits/DataAccessApplicationHelpers.php
```diff
@@ -418,4 +418,19 @@ public function groupApplicationsByProject(LengthAwarePaginator $applications):
...
+    public function returnApplicationsInProject(LengthAwarePaginator $applications): LengthAwarePaginator
+    {
+        $applications->getCollection()->transform(function ($application) {
+            $sameProject = DataAccessApplication::where('project_id', $application['project_id'])
+                ->get();
+            $teams = array();
+            foreach ($sameProject as $app) {
+                $teams = array_merge($teams, $app['teams']->toArray());
+            }
+            $application['teams'] = $teams;
```
calmacx, line 431: Also to avoid defining arrays, merging, using toArray, you can use something like:
```
   $teams = $sameProject->flatMap(function ($app) {
            return $app->teams;
   })->unique('id')->values(); 
   
   $application->teams = $teams;

```

^^ something like that is marginally more readable IMO

app/Http/Traits/DataAccessApplicationHelpers.php
```diff
@@ -418,4 +418,19 @@ public function groupApplicationsByProject(LengthAwarePaginator $applications):
             $page
         );
     }
+
+    public function returnApplicationsInProject(LengthAwarePaginator $applications): LengthAwarePaginator
+    {
+        $applications->getCollection()->transform(function ($application) {
+            $sameProject = DataAccessApplication::where('project_id', $application['project_id'])
```
calmacx, line 425: Might need `DataAccessApplication::with("teams")`, if you're accessing this in the loop?

calmacx (approved):

app/Http/Traits/DataAccessApplicationHelpers.php
BSnelling: Neat, I suppose that is also more memory efficient, given that it defines fewer temporary arrays?

app/Http/Traits/DataAccessApplicationHelpers.php
BSnelling: `teams` is always appended to the model be default.

PR #1242, opened 2025-05-21
feat(GAT-6927): Feature flagging provider

JamieByrneHDRUK:
## Describe your changes
Provider should only be called during boot, updated logging and adding in retry.

### DIFF
--- a/app/Http/Controllers/Api/V1/FeatureFlagController.php
+++ b/app/Http/Controllers/Api/V1/FeatureFlagController.php
@@ -67,6 +67,9 @@ class FeatureFlagController extends Controller
             Log::warning('Invalid API token', ['provided' => $providedToken]);
             return response()->json(['message' => 'Unauthorized: Invalid token.'], 401);
         }
+        Cache::forget('getAllFlags');
+        Cache::forget('feature_flags');
+
 
         $url = env('FEATURE_FLAGGING_CONFIG_URL');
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/V1/FeatureFlagController.php
```diff
@@ -67,6 +67,9 @@ public function index(Request $request, FeatureFlagManager $flagManager): JsonRe
             Log::warning('Invalid API token', ['provided' => $providedToken]);
             return response()->json(['message' => 'Unauthorized: Invalid token.'], 401);
         }
+        Cache::forget('getAllFlags');
+        Cache::forget('feature_flags');
```
BSnelling, line 71: Why do you need to forget both of these? Are they distinct entries in the cache?

JamieByrneHDRUK: so one is for the get endpoint the other is for the provider when first started. Figured its best to clear both

BSnelling (approved):

dnhdruk (approved):

app/Providers/FeatureServiceProvider.php
```diff
@@ -4,40 +4,32 @@
...
+            if (app()->environment('testing') || !$url) {
+                return;
+            }
 
-        if (is_array($featureFlags)) {
-            $flagManager = app(FeatureFlagManager::class);
-            $flagManager->define($featureFlags);
-        }
+            $featureFlags = Cache::remember('feature_flags', now()->addMinutes(60), function () use ($url) {
+                $res = Http::retry(3, 5000)->get($url);
```
calmacx, line 23: may want these nretries, timeout and 'addMinutes' as config/env vars?

calmacx (approved):

app/Providers/FeatureServiceProvider.php
JamieByrneHDRUK: Maybe, wasnt really a requirement for now. Only adding in as github was failing but we have switched to buckets so should hopefully not be an issue

PR #1280, opened 2025-06-20
fix(GAT-5745): double encoded json

JamieByrneHDRUK:
## Describe your changes
Cannot recreate, suspect it is this method. Added in a test for it

### DIFF
--- a/app/Http/Controllers/Api/V1/IntegrationDatasetController.php
+++ b/app/Http/Controllers/Api/V1/IntegrationDatasetController.php
@@ -1191,7 +1191,7 @@ class IntegrationDatasetController extends Controller
             $metadata = json_decode($metadata, true);
         }
         // Pre-process check for incoming data from a resource that passes strings
-        // when we expect an associative array. GMI passes strings, this
+        // when we expect an associative array. FMA passes strings, this
         // is a safe-guard to ensure execution is unaffected by other data types.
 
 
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Api/V1/IntegrationDatasetController.php
```diff
@@ -1183,28 +1183,38 @@ private function getVersion(int $version)
...
+            unset($metadata['metadata']);
             $metadata = $tmpMetadata;
         }
+
+
+
         return $metadata;
     }
 
+    private function isJsonString($value): bool
```
dnhdruk: This may be included in the `helpers.php` file (https://github.com/HDRUK/gateway-api/blob/dev/app/Helpers/helpers.php).

dnhdruk (approved):

calmacx (approved):
Unsure as suspicious of a problem somewhere else if this is all needed.

Tests look good and show the code extractMetadata works as expected, so approving

app/Http/Controllers/Api/V1/IntegrationDatasetController.php
JamieByrneHDRUK: Good shout! moved over cheers

JamieByrneHDRUK:
> Unsure as suspicious of a problem somewhere else if this is all needed.
> 
> Tests look good and show the code extractMetadata works as expected, so approving

Heres a shout, when we get logging in, how about if we meet the conditional we log it out so we can see what the user did when and where the request came from

PR #1281, opened 2025-06-20
feat(GAT-7327): Add V2 endpoints for Data Use Register

spco:
## Describe your changes
Adds V2 `DurController` and `TeamDurController` including testing

Also:
- Remove `technicalSummary` from places it's not needed in V2 DUR endpoints
- Remove `V2/DatasetController::count()` - use the Team version only
- `TeamDatasetController::show()` is no longer restricted to only active datasets
- Fix behaviour of `TeamToolController::update()` so that it is a true PUT rather than PATCH
- Change `ToolController::show()` to  `ToolController::showActive()` so that it only shows active
- Change `DatasetController::show()` to  `DatasetController::showActive()` so that it only shows active
- Don't return team-owned Collections in `V2/UserCollectionController:: indexUserCollection()` (spotted by  @philreekshdr)
- Slim down 3 functions in `ToolsV2Helper` into one with optional args.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7327


app/Http/Traits/DurV2Helpers.php
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    /**
+     * Insert data into ToolHasTag
+     *
+     * @param array $tags
+     * @param integer $toolId
+     * @return mixed
+     */
+    public function insertToolHasTag(array $tags, int $toolId): mixed
```
BSnelling: Does this belong in `DurV2Helpers`?  Could be in an equivalent `Tools` file.  Same comment applies for some other functions in this file.

BSnelling (approved):

app/Http/Traits/DurV2Helpers.php
spco: Good spot

spco: Fixed

JamieByrneHDRUK (approved):

PR #1282, opened 2025-06-20
feat(GAT-7233): Merge logic from Integration Dataset Controller into Dataset Controller

BSnelling:
## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-7233

### DIFF
--- a/app/Http/Traits/CheckAccess.php
+++ b/app/Http/Traits/CheckAccess.php
@@ -3,8 +3,10 @@
 namespace App\Http\Traits;
 
 use Exception;
+use App\Models\Application;
 use App\Models\TeamHasUser;
 use App\Exceptions\UnauthorizedException;
+use Illuminate\Http\Request;
 
 trait CheckAccess
 {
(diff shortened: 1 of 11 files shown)


tests/Feature/DatasetIntegrationDeprecatedTest.php
```diff
@@ -0,0 +1,521 @@
+<?php
```
BSnelling, line 1: This file is a copy of the previous `DatasetIntegrationTest.php` file, so that the soon-to-be-deprecated `v1/integrations/datasets` endpoints are still tested.

tests/Feature/DatasetIntegrationTest.php
spco, line 1: If this is being replaced with *DeprecatedTest, should this file be removed?

app/Http/Traits/CheckAccess.php
```diff
@@ -21,8 +23,24 @@ public function checkAccess(
...
+
+            if ($application) {
+                if (($dbTeamId) && ($dbTeamId !== $application->team_id)) {
+                    throw new UnauthorizedException(
+                        'This Application is not allowed to interact with datasets from another team!'
+                    );
+                }
+                return true;
+            }
+        }
```
spco, line 42: So if `x-application-id` and `x-client-id` are provided but not valid for any Application, this will fall through and attempt to just be the user from JWT. Just checking that is what we want - we don't want to explicitly return false/error if the headers are provided but invalid? Or is this because you're happy to rely on `JwtMiddleware` coming in first?

spco (approved):

app/Http/Traits/CheckAccess.php
BSnelling: If the headers are provided but not valid, it should fail well before this function is hit.  There is a check in JwtMiddleware that will throw a `No known integration matches the credentials provided` error.

tests/Feature/DatasetIntegrationTest.php
BSnelling: We need both, this file tests the `v1/datasets` endpoints being used by an app. *DeprecatedTest  tests the old `v1/integrations/datasets` endpoints. We can delete the *DeprecatedTest file in 6 months time.

dnhdruk (approved):

PR #1293, opened 2025-06-26
fix(GAT-7522): apply restrictions to cohort request endpoints

spco:
## Describe your changes
Lock down several cohort_requests endpoints so that only cohort admin can use them, and lock down the endpoint for a given user's cohort rewuest so only the user can use it.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7522

### DIFF
--- a/app/Http/Controllers/Api/V1/CohortRequestController.php
+++ b/app/Http/Controllers/Api/V1/CohortRequestController.php
@@ -1303,8 +1303,8 @@ class CohortRequestController extends Controller
         $input = $request->all();
         $jwtUser = array_key_exists('jwt_user', $input) ? $input['jwt_user'] : [];
 
-        // Check that either the user has the cohort.read permission, or is asking only for their own record.
-        if (!(($jwtUser['id'] === $id) || in_array("cohort.read", $input['jwt_user']['role_perms']['extra']['perms']))) {
+        // Check that the user is asking only for their own record.
+        if (!($jwtUser['id'] === $id)) {
             throw new UnauthorizedException();
         }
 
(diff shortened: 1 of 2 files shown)


BSnelling (approved):

app/Http/Controllers/Api/V1/CohortRequestController.php
```diff
@@ -1301,6 +1302,12 @@ public function byUser(Request $request, int $id): JsonResponse
     {
         $input = $request->all();
         $jwtUser = array_key_exists('jwt_user', $input) ? $input['jwt_user'] : [];
+
+        // Check that either the user has the cohort.read permission, or is asking only for their own record.
+        if (!(($jwtUser['id'] === $id) || in_array("cohort.read", $input['jwt_user']['role_perms']['extra']['perms']))) {
```
dnhdruk: Do you need?
You have a middleware before that that checks the JWT user permissions.

spco: Yes because I don't want to block someone accessing their own record - so the middleware lets it through to here, where we can check for permissions OR for the requested id matching the jwt id.

dnhdruk: In this case, this permission should apply to all users; however, there are better ways to resolve the situation.

spco: Tbh in the FE the cohort admin does need access to this endpoint - only the user does - but it seemed appropriate to let the cohort admin also access this for convenience.

I could remove the perms check, and that leaves just the check that the requested user id matches the current user id. I could then simply use `$this->checkAccess($input, null, $id, 'user')`.

Would you prefer that?

dnhdruk: Please keep it simple. This endpoint only requires the user ID & jwt.

spco: Done. Please re-review 👍

dnhdruk (approved):

PR #1306, opened 2025-07-04
fix(GAT-7457): Prework to fix pipeline and phpstan errors

loki-sinclair-hdruk:
## Describe your changes
Couldn't run the pipeline from a fresh install. This fixes that, and also addresses all of the phpstan errors I was seeing.

### DIFF
--- a/app/Providers/HelperServiceProvider.php
+++ b/app/Providers/HelperServiceProvider.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace App\Providers;
+
+use Illuminate\Support\ServiceProvider;
+
+class HelperServiceProvider extends ServiceProvider
+{
+    /**
+     * Register services.
+     */
+    public function register(): void
+    {
+        //
+    }
+
+    /**
+     * Bootstrap services.
+     */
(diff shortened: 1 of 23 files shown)


calmacx (approved):

app/Providers/HelperServiceProvider.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
```
dnhdruk, line 2: link for `app/Helpers/helpers.php` is already in `https://github.com/HDRUK/gateway-api/blob/649e59ec123174bfa8be3287db792328347f5b4b/composer.json#L76`.
do you need this ?

loki-sinclair-hdruk: Yes, because at the point of `db:seed` composer isn't autoloaded. If you completely wipe your clone, and start fresh you cannot migrate/seed from a fresh clone because of the usage of functions that haven't been loaded because app hasn't been bootstrapped in the same way.

dnhdruk (approved):

spco (approved):

PR #1308, opened 2025-07-07
chore(GAT-7451): CE changes. Removes sole SU and replaces with named officials.

loki-sinclair-hdruk:
## Describe your changes
Retires the developers@ account and replaces with individual role based perms. Also adds in a command for easier use in the future.

For the command usage:

```
php artisan app:super-user-controls <email address> --action [add|remove]
```

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7451

No

Yes


app/Console/Commands/SuperUserControls.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $email = $this->argument('email');
+        $action = $this->option('action');
+
+        $user = User::where('email', $email)->select('id', 'email')->first();
```
dnhdruk, line 33: `is_admin=1`

loki-sinclair-hdruk: Explain more?

dnhdruk: super user / super admin = `users.is_admin = 1 `

spco:
@loki-sinclair-hdruk Can I check what is the suggested flow we should use for local/dev testing of unprivileged accounts? I often use my hdr account to test as a "random" unprivileged user, but this change will mean my go-to account now has superuser access.

On local I can always tweak permissions/roles in the DB, but is there a better way we could support this, particularly on dev? Or suggestions how I should change my own flow?

app/Console/Commands/SuperUserControls.php
loki-sinclair-hdruk: Eh? Since when does the db flag override the role?

loki-sinclair-hdruk:
Partly why I included the command, so things can be changed on a whim as needed.

You're right in that local/dev is subject to updating the db. But in order to satisfy CE, we have to demonstrate sensible locking down. There were a number of duplicate email addresses for individuals in the database (prod) - and I opted for hdruk.ac.uk ones to be the privileged ones to be consistent with business policy. This was cleared with Stephen. I didn't have to include everywhere for each env, but I do love consistency!

app/Console/Commands/SuperUserControls.php
dnhdruk: A super-admin user has been defined as "users.is_admin=1"
example:
```
https://github.com/HDRUK/gateway-api/blob/dev/app/Http/Middleware/CheckAccessMiddleware.php#L59
```

dnhdruk: The initial deployment was created command:
```
https://github.com/HDRUK/gateway-api/blob/dev/app/Console/Commands/AddSuperAdminToAllTeams.php
```

loki-sinclair-hdruk: Resolved in last update

dnhdruk (approved):
We need to test it to determine what changes we can make in other areas.

spco:
> Partly why I included the command, so things can be changed on a whim as needed.
> 
> You're right in that local/dev is subject to updating the db. But in order to satisfy CE, we have to demonstrate sensible locking down. There were a number of duplicate email addresses for individuals in the database (prod) - and I opted for hdruk.ac.uk ones to be the privileged ones to be consistent with business policy. This was cleared with Stephen. I didn't have to include everywhere for each env, but I do love consistency!

That's fair enough - for local, the command does make it super easy, now you point it out.

calmacx (approved):

PR #1310, opened 2025-07-09
feat(GAT-5632): Change containsTissue name for FE chip updating

tmillingtonhic:
## Screenshots (if relevant)
Screenshot 2025-07-09 at 18 30 12

## Describe your changes

Adds a command to update the filters in the MySQL DB (app:update-filters-dataset-gat5632), updates the search controller with the new filter name and updates Elastic

FE: https://github.com/HDRUK/gateway-web/pull/1255

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-5632

Requires app:update-filters-dataset-gat5632 and an elastic re-index

### DIFF
--- a/app/Console/Commands/UpdateFiltersDatasetGat5632.php
+++ b/app/Console/Commands/UpdateFiltersDatasetGat5632.php
@@ -5,7 +5,7 @@ namespace App\Console\Commands;
 use Illuminate\Console\Command;
 use App\Models\Filter;
 
-class updateFiltersDatasetGat5632 extends Command
+class UpdateFiltersDatasetGat5632 extends Command
 {
     /**
      * The name and signature of the console command.
--- a/app/Console/Commands/updateFiltersDatasetGat5632.php
+++ b/app/Console/Commands/updateFiltersDatasetGat5632.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use App\Models\Filter;
+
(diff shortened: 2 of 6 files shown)


dnhdruk (approved):

app/Console/Commands/UpdateFiltersDatasetGat5632.php
```diff
@@ -0,0 +1,37 @@
...
+    public function handle()
+    {
+        Filter::updateOrCreate([
+            'type' => 'dataset',
+            'keys' => 'containsTissue',
+        ], [
+            'type' => 'dataset',
+            'keys' => 'containsBioSamples',
+        ]);
+    }
```
calmacx, line 36: feels like if this is changing filters - there needs to be something to trigger a re-index of anything this effects?

Been a while, so would check with @BSnelling @spco

calmacx (commented):
Changing filters is always dangerous - so unsure about approving without checking with the elastic experts

app/Console/Commands/UpdateFiltersDatasetGat5632.php
BSnelling: There will need to be a reindex of elastic datasets for this to take effect, so it is worth double checking that that is noted on the next release plan as a manual step @tmillingtonhic. 

Otherwise this should be fine.  Since the data type is a boolean we shouldn't need to do anything additional on the search service side.

BSnelling (approved):

PR #1324, opened 2025-07-16
feat(GAT-7329): Add dataset linkage information to payload for display on landing page.

spco:
## Describe your changes
This was missing from the v2 payload, but is now present.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7329


BSnelling (approved):

calmacx (approved):

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -856,4 +859,71 @@ public function exportMock(Request $request)
             throw new Exception($e->getMessage());
         }
     }
+
+    public function getLinkages($datasetVersionId)
+    {
+        $datasetLinkages = DatasetVersionHasDatasetVersion::where([
```
dnhdruk, line 865: If "DatasetVersionHasDatasetVersion" is updated when the dataset state changes to "ACTIVE," this method would be less expensive.

spco: Most linkage tables are symmetric, so the typical V2 paradigm is to leave the linkage in place regardless of the status of either entity, and then check their statuses at time of lookup.

But you're right that in this special case, the linkage is not symmetric. I'll put in a ticket to explore optimising this by using an observer to update a status column in this table that matches the target DatasetVersions's Dataset's status column. Too big a piece of work for now. 👍

PR #1327, opened 2025-07-17
chore(GAT-7294): Add contextual logging

BSnelling:
## Describe your changes

Adds logging with context in key areas:
- middleware to log incoming request and the response (I lifted that middleware from the registry - thank you registry!)
- in dispatched jobs - so we can track the requests that initiated those jobs
- wherever we call out to other services so that we can pass the `'x-request-session-id'` on

Note: when we upgrade to Laravel 11/12 we can use `Context` for this logging which will be more elegant (https://laravel.com/docs/12.x/context#main-content), but it is not available in Laravel 10.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-7294

### DIFF
--- a/app/Http/Middleware/LogRequestResponse.php
+++ b/app/Http/Middleware/LogRequestResponse.php
@@ -22,7 +22,7 @@ class LogRequestResponse
 
         // Log the incoming request
         Log::info('Request', [
-            'request_id' => $request->headers->all()['x-request-session-id'] ?? null,
+            'x-request-session-id' => $request->headers->all()['x-request-session-id'] ?? null,
             'url' => $request->fullUrl(),
             'http_method' => $request->getMethod(),
             'method_name' => $methodName,
@@ -37,7 +37,7 @@ class LogRequestResponse
         $responseBody = $this->maskSensitive($responseBody);
         $responseBody = $this->truncateBody($responseBody);
         Log::info('Response', [
-            'request_id' => $request->headers->all()['x-request-session-id'] ?? null,
+            'x-request-session-id' => $request->headers->all()['x-request-session-id'] ?? null,
             'url' => $request->fullUrl(),
             'http_method' => $request->getMethod(),
             'method_name' => $methodName,
--- a/app/Http/Middleware/LogRequestResponse.php
+++ b/app/Http/Middleware/LogRequestResponse.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Log;
+
+class LogRequestResponse
+{
+    // List of sensitive fields to mask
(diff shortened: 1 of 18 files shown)


app/Http/Middleware/LogRequestResponse.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function handle(Request $request, Closure $next)
+    {
+        // Mask sensitive fields in request body
+        $body = $this->maskSensitive($request->all());
+        $body = $this->truncateBody($body);
+        $methodName = $this->getMethodName($request);
+
+        // Log the incoming request
+        Log::info('Request', [
```
BSnelling: Made this an `info` level log but should it be `debug` as this could get quite noisy?

app/Http/Middleware/LogRequestResponse.php
```diff
@@ -0,0 +1,84 @@
...
+            'x-request-session-id' => $request->headers->all()['x-request-session-id'] ?? null,
+            'url' => $request->fullUrl(),
+            'http_method' => $request->getMethod(),
+            'method_name' => $methodName,
+            'body' => $body,
```
BSnelling, line 29: Do we want to log the body? Even truncated that will be a lot of content.

app/Http/Middleware/LogRequestResponse.php
```diff
@@ -0,0 +1,84 @@
...
+        ]);
+
+        $response = $next($request);
+
+        // Log the outgoing response
+        $responseBody = method_exists($response, 'getContent') ? $response->getContent() : null;
+        $responseBody = json_decode($responseBody, true);
+        $responseBody = $this->maskSensitive($responseBody);
+        $responseBody = $this->truncateBody($responseBody);
+        Log::info('Response', [
```
dnhdruk: Can you add request and response under one single `Log::info`?

dnhdruk: The middleware is sufficient; there is no need to update the controllers, as all the necessary data is already available here.

app/Http/Middleware/LogRequestResponse.php
soringumeni1: Indeed, if something is too noisy, we should make it debug. We can enable it if required.

soringumeni1 (approved):

app/Http/Middleware/LogRequestResponse.php
BSnelling: Changed this middleware to `debug` level.  The dev and preprod environments have logging at to `debug` anyway so we'll see them there.

app/Http/Middleware/LogRequestResponse.php
BSnelling: I updated the controllers which were going to call out to another service. Otherwise leaving the logging to this middleware for now.

dnhdruk (approved):

PR #1368, opened 2025-08-18
feat(GAT-1234): Change collection_has_dataset_version table primary key

spco:
... to an int rather than a composite primary key.

## Describe your changes
As per title

Database migration

### DIFF
--- a/database/migrations/2025_08_15_091200_add_primary_key_to_collection_has_dv_table.php
+++ b/database/migrations/2025_08_15_091200_add_primary_key_to_collection_has_dv_table.php
@@ -7,11 +7,20 @@ use Illuminate\Support\Facades\Schema;
 return new class () extends Migration {
     public function up(): void
     {
+        if (DB::getDriverName() !== 'sqlite') {
+            Schema::table('collection_has_dataset_version', function (Blueprint $table) {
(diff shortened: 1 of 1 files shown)


database/migrations/2025_08_15_091200_add_primary_key_to_collection_has_dv_table.php
```diff
@@ -0,0 +1,118 @@
...
+        });
+
+        DB::table('collection_has_dataset_version')->insert(
+            DB::table('collection_has_dataset_version_backup')
+                ->select('collection_id', 'dataset_version_id', 'user_id', 'application_id', 'reason', 'created_at', 'updated_at', 'deleted_at')
+                ->get()
+                ->map(function ($row) {
+                    return (array) $row;
+                })
+                ->toArray()
```
calmacx, line 58: needs to be chunked? could be loading up a lot to memory here if there are 10,000s of rows?

spco: There's no more than 1500 rows right now, so I think that's acceptable without chunking - unless you disagree?

calmacx (approved):

loki-sinclair-hdruk (approved):

PR #1370, opened 2025-08-18
fix(GAT-7678): Brings GMI into API to reduce significant complexity

loki-sinclair-hdruk:
## Screenshots (if relevant)
image

## Describe your changes
So, this is a massive change. It basically supersedes GMI as a service, and brings it into the API. I've pulled all active federations by run time, and then spawn a job for each to run concurrently and left-shifted. As far as I can tell, it creates/updates and deletes what is necessary per each run. The attached screenshot displays a re-run against MDW/KMS that detects no changes are required.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7678

Yes:

```
GOOGLE_APPLICATION_CREDENTIALS="" // Needs to be a full path to the credentials json file for gcloud
GOOGLE_APPLICATION_PROJECT_PATH="" // Needs to be the project ID within GCP
GATEWAY_API_USER_ID= // Needs to be the service layer user
```

No.


app/Traits/GatewayMetadataIngestionTrait.php
```diff
@@ -0,0 +1,245 @@
...
+    ): void {
+        $toCreate = $remoteItems->keys()->diff($localItems->keys());
+        foreach ($toCreate as $pid) {
+            $data = $remoteItems[$pid];
+            $response = Http::get($this->makeDatasetUrl($federation, $data), $this->determineAuthType($federation, $gms));
+            if ($response->status() === 200) {
+                try {
+                    $input = [
+                        'status' => 'ACTIVE',
+                        'create_origin' => 'GMI',
```
JamieByrneHDRUK, line 118: Didn't we change references from GMI to FMA?

JamieByrneHDRUK (approved):

app/Traits/GatewayMetadataIngestionTrait.php
loki-sinclair-hdruk: Other way around pal. FMA -> GMI.

app/Traits/GatewayMetadataIngestionTrait.php
```diff
@@ -0,0 +1,245 @@
...
+                if ($response->status() === 200) {
+                    $team = Team::where('id', $gmi->getTeam())->first();
+                    $ds = Dataset::where('pid', $pid)->first();
+                    $dv = DatasetVersion::where('dataset_id', $local->id)->orderBy('id', 'desc')->first()->toArray();
+
+                    $payload = [
+                        'extra' => [
+                            'id' => $ds->id,
+                            'pid' => $ds->pid,
+                            'datasetType' => 'Health and disease',
```
calmacx, line 157: not sure about the hardcoding of this - would be model dependent - but presumably this is a copy&paste over from the old GO code

calmacx (approved):

app/Traits/GatewayMetadataIngestionTrait.php
loki-sinclair-hdruk: Yeah, I noticed this at the time, but very definitely what is in current Go implementation and thus carried forward. This is also muddied with potential subTypes being added in v4, but that along with this will be for future.

PR #1392, opened 2025-09-02
fix(GAT-7678): Fix for MDW East mids which didn't return a fully formed JSON response, and thus failed pull. Changes work for both KMS and EM.

loki-sinclair-hdruk:
## Describe your changes
It was noted by MDW that the GMI service was unable to pull listings from East Mids integration. This was debugged to expose that Http::get call to `->json()` returned null on _only_ EM datasets. KMS remained unaffected. I've updated the code to use `->body()` from the response object, and do the `json_decode(...)` ourselves, which seems to be slightly more resilient when faced with an enigma of badly formed json.

### DIFF
--- a/app/Traits/GatewayMetadataIngestionTrait.php
+++ b/app/Traits/GatewayMetadataIngestionTrait.php
@@ -4,6 +4,8 @@ namespace App\Traits;
 
 use Http;
 use Config;
+use MetadataManagementController as MMC;
+
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Collection;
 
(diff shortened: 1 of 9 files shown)


app/Traits/GatewayMetadataIngestionTrait.php
```diff
@@ -77,27 +82,42 @@ public function getLocalDatasetsForFederatedTeam(GatewayMetadataIngestionService
         ])->get())->keyBy('pid');
     }
 
-    public function deleteLocalDatasetsNotInRemoteCatalogue(Collection $localItems, Collection $remoteItems): void
+    public function deleteLocalDatasetsNotInRemoteCatalogue(Collection $localItems, Collection $remoteItems): bool
     {
+        $this->log('info', 'testing REMOTE collection for LOCAL deletions');
```
spco, line 87: Just checking, the use of 'testing' here is to mean "checking", and this it to be kept, it's not debug logging that should be removed?

spco (approved):

app/Traits/GatewayMetadataIngestionTrait.php
spco: No need to change it if so, just clarifying for my understanding of intent

dnhdruk (approved):

app/Traits/GatewayMetadataIngestionTrait.php
loki-sinclair-hdruk: Aye, correct. Possibly a bad use of wording, but you understood it.

PR #1413, opened 2025-09-19
fix!(GAT-7759): Added sanitation on the search term and filters to search requests

tmillingtonhic:
## Screenshots (if relevant)
https://github.com/user-attachments/assets/3bab34b1-e051-4958-9a27-85cf255cd30d

## Describe your changes
Search terms are now stripped of non alphanumeric or - and _ characters
Filters can now only be set to specific values
x-header-request-id is now validated to only be alphanumeric plus : and -

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7759
https://hdruk.atlassian.net/browse/GAT-7760

### DIFF
--- a/app/Http/Controllers/Api/V1/SearchController.php
+++ b/app/Http/Controllers/Api/V1/SearchController.php
@@ -53,6 +53,7 @@ use App\Http\Requests\Search\PublicationSearch;
 use Illuminate\Http\Client\ConnectionException;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use App\Http\Requests\Search\Search;
+
 class SearchController extends Controller
 {
     use IndexElastic;
--- a/app/Http/Controllers/Api/V1/SearchController.php
+++ b/app/Http/Controllers/Api/V1/SearchController.php
@@ -49,11 +49,10 @@ use App\Models\ToolHasProgrammingPackage;
 use App\Models\ToolHasProgrammingLanguage;
 use App\Http\Traits\GetValueByPossibleKeys;
 use App\Models\PublicationHasDatasetVersion;
-use Illuminate\Database\Eloquent\Casts\Json;
 use App\Http\Requests\Search\PublicationSearch;
 use Illuminate\Http\Client\ConnectionException;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
-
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -49,11 +49,10 @@
 use App\Models\ToolHasProgrammingLanguage;
 use App\Http\Traits\GetValueByPossibleKeys;
 use App\Models\PublicationHasDatasetVersion;
-use Illuminate\Database\Eloquent\Casts\Json;
 use App\Http\Requests\Search\PublicationSearch;
 use Illuminate\Http\Client\ConnectionException;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
-
+use App\Http\Requests\Search\Search;
```
spco, line 55: Please reinstate the new line - have you run this through a linter?

spco (changes requested):
I'm not sure about this as a solution. For a free-text search, I think users will dislike being told they're using an illegal character - much better to silently sanitize the string rather than reject it. There are all sorts of times when a user would want to put in a `(` or an `&` etc - that's a legitimate part of the titles of many datasets and entities - so we shouldn't reject them but instead sanitize them.

Have we established whether there's a real pathway to evaluating injected SQL code? This feels like we're patching a hole that doesn't really exist. Good to have defense in depth, but this will irritate users.

tmillingtonhic:
Hmm maybe I'm misunderstanding what needs to be resolved then - I thought the requirement was the API should reject requests with invalid parameters. I did think it would be annoying - could probs do some FE validation on that though. I did check with Loki as I'm not sure this is actually a SQL injection issue, I think the tool is just picking that if you have a large number of results returned the results will differ, but he said it was worth patching.

spco:
Ok yep. No issue with this being patched, but I don't think this can go to production as-is - perhaps check with Loki but I'd imagine we can do simple validation (is this a string, is it not vastly long) in the validation phase and then beef up our sanitization stage at the beginning of our endpoints. Or perhaps as in https://laravel.com/docs/12.x/validation#preparing-input-for-validation ?

spco (approved):

dnhdruk (approved):

PR #1442, opened 2025-10-22
feat(GAT-7803): upgrade Laravel to v12

spco:
## Describe your changes
Upgrade to Laravel v12. This required:

- Tweak Carbon functions now we use Carbon v3 rather than v2.
- Add a load of type-hinting for larastan to be satisfied.
- The main killer change here is that we can no longer call `env(BLAH)` outside of the `config/` folder. Instead, all env variables must be read in at start time, then accessed via `config(BLAHBLAH)`. This includes changing `env(GATEWAY_URL)` to `config(gateway.gateway_url)` in a hell of a lot of files for email templates. Includes a migration that will update all existing templates.
- Also move loads of bits of code (getting user values) out of try-catch since their results is used in the catch (for larastan)

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7803

Yes, one


app/Http/Controllers/Api/V2/DurController.php
```diff
@@ -187,16 +187,7 @@ public function indexActive(Request $request): JsonResponse
                 ])
             )
             ->applySorting()
-            ->paginate((int) $perPage, ['*'], 'page')
-            ->through(function ($dur) {
-                if ($dur->datasets) {
```
loki-sinclair-hdruk, line 192: Can understand paginate, but do we no longer need this filtering?

loki-sinclair-hdruk:
Presume that we've tested an email send, just to ensure the URL works under the config(...) change too.

app/Http/Controllers/Api/V2/DurController.php
spco: This isn't filtering, it would be modifying `dur->datasets` to add the title to each dataset.

However, we don't make use of this anywhere, and more importantly, larastan has pointed out that the `dur->datasets` field doesn't actually exist! So I've just removed it given it's unused anyway.

spco:
> Presume that we've tested an email send, just to ensure the URL works under the config(...) change too.

Yes (locally) - took a bit of figuring out as it wasn't at first. The migration tweaks all templates so they use `config(gateway.gateway_url)` instead of `env(GATEWAY_URL)`, and the text-replacement function's been changed to handle that tweaked format. So all good 👍

loki-sinclair-hdruk (approved):

JamieByrneHDRUK (approved):

PR #1470, opened 2025-11-21
fix(GAT-8025): Unstable Collections form page and unable to add datasets

tmillingtonhic:
## Screenshots (if relevant)

https://github.com/user-attachments/assets/ad932191-0766-4001-bb2e-42a8e7223681

https://github.com/user-attachments/assets/8008d743-c285-4d9b-a82c-2166e6c27746

## Describe your changes
Batched CollectionHasDatasetVersion inserts and deletes so there's only one elastic reindex

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-8025


dnhdruk (approved):

loki-sinclair-hdruk:
I'm not sure this necessarily makes indexing any less intrusive, as the indexes are run on the model being changed, via Observer. Do you have any evidence/proof that this combats what you think it does? More than happy to be proven wrong! 😄

tmillingtonhic:
> I'm not sure this necessarily makes indexing any less intrusive, as the indexes are run on the model being changed, via Observer. Do you have any evidence/proof that this combats what you think it does? More than happy to be proven wrong! 😄

So I thought that by batching the requests on the model that would only run the re-index once per request rather than on every loop? It does run a lot more quickly with this change - before you couldn't have more than 5(ish) datasets before it would fall over.

spco:
I'd add that we need to double-check that the Observer even fires at all. I know that if you do a batch-update e.g.
```
$colls = Collection::where('team_id' => 3)->get();
$colls->update('status' => 'ACTIVE');
```
(I haven't checked that's valid PHP!) then, because it's not a single update e.g. with `->first()`, the `updated` Observer doesn't fire. https://laravel.com/docs/12.x/eloquent#mass-updates

The documentation isn't totally clear about what would happen in this case where you're using `insert` rather than `create`. But it needs checking rather than assuming - you may simply not be triggering the reindexing at all!

calmacx (approved):

PR #1489, opened 2025-12-22
fix(GAT-7839): Follow up can't be > 10 years due to HTML encoding

tmillingtonhic:
## Screenshots (if relevant)
https://github.com/user-attachments/assets/8269e7da-b28b-4b37-a18e-3127b9100bcd

## Describe your changes
Laravel was encoding > to &gt;, which traser was then rejecting. > is now allowed in the metadata to solve this

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-7839

### DIFF
--- a/app/Http/Traits/DatasetsV2Helpers.php
+++ b/app/Http/Traits/DatasetsV2Helpers.php
@@ -123,7 +123,6 @@ trait DatasetsV2Helpers
             $metadata = $tmpMetadata;
         }
         if (is_array($metadata) && array_key_exists("metadata", $metadata)) {
-            // Remove HTML encoding from fields that will be validated by traser
             $this->decodeMetadataHTML($metadata['metadata']);
         }
 
@@ -132,6 +131,8 @@ trait DatasetsV2Helpers
 
     private function decodeMetadataHTML(array &$metadataArr)
     {
+        // coverage.followUp can contain >, which will be encoded by the BE
+        // traser will reject this so we need to decode it
         if (Arr::has($metadataArr, 'coverage.followUp')) {
             $metadataArr['coverage']['followUp'] = html_entity_decode($metadataArr['coverage']['followUp']);
         }
--- a/app/Http/Traits/DatasetsV2Helpers.php
+++ b/app/Http/Traits/DatasetsV2Helpers.php
@@ -123,7 +123,18 @@ trait DatasetsV2Helpers
             $metadata = $tmpMetadata;
         }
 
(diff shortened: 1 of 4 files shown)


dnhdruk (approved):

app/Http/Traits/DatasetsV2Helpers.php
```diff
@@ -122,8 +122,19 @@ private function extractMetadata(Mixed $metadata)
...
 
         return $metadata;
     }
 
+    private function decodeMetadataHTML(array &$metadataArr)
+    {
+        // coverage.followUp can contain >, which will be encoded by the BE
+        // traser will reject this so we need to decode it
+        if (Arr::has($metadataArr, 'coverage.followUp')) {
+            $metadataArr['coverage']['followUp'] = html_entity_decode($metadataArr['coverage']['followUp']);
```
calmacx: What is metadataArr here? GWDM? Which version? Will `['coverage']['followUp']` always be present/in this location? If the GWDM changes then this code has to be changed too? 

Why does TRASER reject this? Traser rejecting it or the schema is rejecting?

tmillingtonhic: I think it'd be GWDM, but good point. This would have to be changed if GWDM was changed.

The issue is that one of the middleware encodes any HTML characters it comes across. Follow up can be` > 10 years`, which is then encoded as  `&gt; 10 years.`

 This isn't a valid value as far as the schema is concerned, so it gets rejected.

calmacx: Ah i see. 
```
class FollowupV2(Enum):
   ...
   field__10_YEARS = '> 10 Years'
```
And the middleware is sanitising JSON data to prevent against any injection attacks. 

I think the solution to this problem needs to be elsewhere, otherwise we'll end up with code full of patches. 

The sanitising is to prevent injection attacks saved to the SQL database (and this metadata would be saved). Traser is public and doesn't talk to any critical infrastructure, so middleware could be added in that to safely re-encode data it receives before it validates the payload against the schema (or maybe when the data is posted to TRASER it is re-encoded at that point?) ?

tmillingtonhic: I think that might be more sensible, I did look at just removing the sanitation for inputs into traser but that did make the system vulnerable to injection as traser doesn't do much validation on large text fields. So ended up back on patching individual fields...

calmacx (approved):

JamieByrneHDRUK (approved):

PR #1493, opened 2026-01-05
fix(GAT-8219): Running out of memory when fetching large collections

tmillingtonhic:
## Screenshots (if relevant)
Screenshot 2026-01-05 at 14 50 49

https://github.com/user-attachments/assets/1c286112-f8a8-4b3c-8da4-ece4d0449f09

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-8219

### DIFF
--- a/app/Http/Controllers/Api/V2/TeamCollectionController.php
+++ b/app/Http/Controllers/Api/V2/TeamCollectionController.php
@@ -752,7 +752,7 @@ class TeamCollectionController extends Controller
             } else {
                 $this->deleteCollectionFromElastic((int) $id);
             }
-            
+
             Auditor::log([
                 'user_id' => $userId,
                 'target_team_id' => array_key_exists('team_id', $array) ? $array['team_id'] : null,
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/V2/TeamCollectionController.php
```diff
@@ -955,9 +958,11 @@ public function edit(EditTeamCollection $request, int $teamId, int $id): JsonRes
                 'description' => 'Collection ' . $id . ' updated',
             ]);
 
-            return response()->json([
+            $collection = $this->getCollectionActiveById($id);
```
loki-sinclair-hdruk: If we're now streaming the collection data, why do we need the additional memory to store as a variable? Was the original call to the function problematic?

loki-sinclair-hdruk: I'd hazard a guess that we're actually using more memory here, just masking with the stream

tmillingtonhic: The issue is in returning the json - Laravel uses a large amount of memory constructing the response()->json(). Without calling the response()->json() method it only takes about 30MB to construct the collection. By streaming the json it means we don't have to construct the full json response before returning. 

Rewriting `$this->getCollectionActiveById` to stream the results back would probably reduce memory consumption further but just streaming the json was sufficient for this ticket.

loki-sinclair-hdruk (approved):

dnhdruk (approved):

PR #1504, opened 2026-01-28
fix(GAT-0000): Fix issue with data access seeder

philreekshdr:
## Describe your changes
Fix issue with data access seeder

When running the seeder on a fresh db, errors with 

```
Cannot truncate a table referenced in a foreign key constraint (`db`.`dar_template_has_files`, CONSTRAINT `dar_template_has_files_template_id_foreign`) (Connection: mysql, SQL: truncate table `dar_templates`)
```

using query delete, ignores the fk

### DIFF
--- a/database/seeders/DataAccessTemplateSeeder.php
+++ b/database/seeders/DataAccessTemplateSeeder.php
@@ -6,6 +6,7 @@ use App\Models\DataAccessTemplate;
 use App\Models\DataAccessTemplateHasQuestion;
 use App\Models\QuestionBank;
 use Illuminate\Database\Seeder;
+use Illuminate\Support\Facades\DB;
 
 class DataAccessTemplateSeeder extends Seeder
 {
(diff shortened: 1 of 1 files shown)


database/seeders/DataAccessTemplateSeeder.php
```diff
@@ -14,8 +14,8 @@ class DataAccessTemplateSeeder extends Seeder
      */
     public function run(): void
     {
-        DataAccessTemplate::truncate();
-        DataAccessTemplateHasQuestion::truncate();
+        DataAccessTemplate::query()->delete();
+        DataAccessTemplateHasQuestion::query()->delete();
```
calmacx: erm probably does work, but usually the done thing is to wrap in:

```
 DB::statement('SET FOREIGN_KEY_CHECKS=0');
 DataAccessTemplate::truncate();
 DataAccessTemplateHasQuestion::truncate(); 
 DB::statement('SET FOREIGN_KEY_CHECKS=1');
```

calmacx (approved):
if doesnt work, use suggestion in comments

database/seeders/DataAccessTemplateSeeder.php
calmacx: your one will work if we are not using soft-deletes

loki-sinclair-hdruk: Yep, this is the usual approach. `->delete()` as Calum says will try to use `deleted_at` first, which won't prevent duplicate throws.

Looking at the models, DataAccessTemplate makes use of SoftDeletes, so this won't work. Needs a more controlled, but heavier hand @philreekshdr.

dnhdruk: I think is like this
```
DB::statement('SET FOREIGN_KEY_CHECKS=0');
 DataAccessTemplateHasQuestion::truncate(); 
 DataAccessTemplate::truncate();
 DB::statement('SET FOREIGN_KEY_CHECKS=1');
```
delete the relation and after the origin

philreekshdr: cheers, updated 

- just saw the extra review comments

dnhdruk (approved):

PR #1512, opened 2026-02-04
fix(GAT-8009): Data issue causing bug when adding Dataset

dnhdruk:

### DIFF
--- a/app/Http/Controllers/Api/V1/CohortRequestController.php
+++ b/app/Http/Controllers/Api/V1/CohortRequestController.php
@@ -2,32 +2,33 @@
 
 namespace App\Http\Controllers\Api\V1;
 
-use Config;
-use Auditor;
-use Exception;
-use App\Models\User;
-use App\Models\OauthUser;
-use App\Jobs\SendEmailJob;
-use App\Models\Permission;
-use Illuminate\Http\Request;
-use App\Models\CohortRequest;
-use App\Models\EmailTemplate;
-use Illuminate\Support\Carbon;
-use App\Models\CohortRequestLog;
-use Illuminate\Http\JsonResponse;
-use App\Models\CohortRequestHasLog;
-use App\Http\Controllers\Controller;
-use App\Http\Traits\HubspotContacts;
 use App\Exceptions\UnauthorizedException;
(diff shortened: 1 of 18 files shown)


app/Models/DatasetVersion.php
```diff
@@ -102,10 +102,7 @@ public function metadata(): Attribute
     */
     public function scopeFilterTitle(Builder $query, string $filterTitle): Builder
     {
-        return $query->whereRaw(
-            "LOWER(JSON_EXTRACT(JSON_UNQUOTE(metadata), '$.metadata.summary.title')) LIKE LOWER(?)",
-            ["%$filterTitle%"]
-        );
+        return $query->where('title', 'LIKE', "%{$filterTitle}%");
```
loki-sinclair-hdruk, line 105: Was this part of the bug? The raw sql will be quicker, no?

loki-sinclair-hdruk: Ignore this, I see what you're doing below.

loki-sinclair-hdruk:
What was the actual bug though? I'm unsure if duplicating data is needed, and that's what I'm trying to understand.

app/Http/Controllers/Api/V1/CohortRequestController.php
```diff
@@ -1779,7 +1779,7 @@ public function indicateNhseAccess(Request $request, int $id): JsonResponse
 
         // Check that the user is asking only for their own record.
         if (! ($jwtUser['id'] === $id)) {
-            throw new UnauthorizedException;
+            throw new UnauthorizedException();
```
calmacx, line 1782: ? is this our linters messing with each other?

dnhdruk: pure lint

loki-sinclair-hdruk:
Approving under the proviso that we should investigate this as a functional index, rather than duplicate the title. Although, I do see the benefit in having the title elsewhere, but, functional indexes would make create links to it regardless, so solution is debatable.

loki-sinclair-hdruk (approved):

Vikas00230 (approved):

PR #1519, opened 2026-02-11
fix(GAT-8523): Reworking Feature Flags for users

calmacx:
## Describe your changes

* Remove FeatureModels (feature_flags) table 
* New routes for feature flags per user
* route to get feature flag for a user


spco:
Have you spoken to @JamieByrneHDRUK re using Pennant? I have vague memories that the initial few iterations of feature flags used pennant and were unperformant, so we moved to an in-DB solution, but I wasn't close enough to it to know whether Pennant was identified as the true cause.

app/Console/Commands/FeatureSet.php
```diff
@@ -0,0 +1,65 @@
...
+
+        $on ? $store->activate($name) : $store->deactivate($name);
+        Feature::flushCache();
+
+        $state = $store->active($name) ? 'ON' : 'OFF';
+        $this->info("{$name} is now {$state} for {$scopeLabel}");
+
+        return self::SUCCESS;
+    }
+}
```
calmacx, line 65: new command to allow me to set user specific feature flags - we'll use this to enable CDS for some users (we can also do this via the FE though too)

app/Models/Feature.php
```diff
@@ -1,27 +0,0 @@
...
-
-    /**
-     * The attributes that are mass assignable.
-     */
-    protected $fillable = [
-        'name',
-        'scope',
-        'value'
-    ];
-}
```
calmacx, line 27: Not needed - we can use `use Laravel\Pennant\Feature;`

app/Providers/AppServiceProvider.php
```diff
@@ -31,9 +32,9 @@ public function register()
                     }
                 }
 
-                $sql = str_replace(array('%', '?'), array('%%', '%s'), $query->sql);
+                $sql = str_replace(['%', '?'], ['%%', '%s'], $query->sql);
                 $sql = vsprintf($sql, $bindings);
-                \Log::warning("SQL query: " . $sql, ['time' => $query->time]);
+                \Log::warning('SQL query: '.$sql, ['time' => $query->time]);
```
calmacx, line 37: my linter changed this (?)

app/Providers/AppServiceProvider.php
```diff
@@ -63,6 +64,7 @@ public function boot()
         Passport::refreshTokensExpireIn(now()->addDays(30));
 
         DataAccessApplication::observe(DataAccessApplicationObserver::class);
-    }
 
+        Feature::resolveScopeUsing(fn ($driver) => null);
```
calmacx, line 68: make sure the default scope falls back to the global

database/migrations/2026_02_11_091519_drop_feature_flags_table.php
```diff
@@ -0,0 +1,22 @@
...
+
+    public function down(): void
+    {
+        Schema::create('feature_flags', function (Blueprint $table) {
+            $table->id();
+            $table->string('key')->unique();
+            $table->boolean('enabled')->default(false);
+            $table->timestamps();
+        });
+    }
```
calmacx, line 21: get rid of this redundant table

calmacx:
@spco - no i did not as @JamieByrneHDRUK was on AL yesterday. We were using pennant before but we had just switched to `'default' => env('PENNANT_STORE', 'database'),` so I havent change that, it still relies on the DB

app/Http/Controllers/Api/V1/CohortRequestController.php
```diff
@@ -1587,7 +1598,7 @@ public function byUser(Request $request, int $id): JsonResponse
 
         // Check that the user is asking only for their own record.
         if (! ($jwtUser['id'] === $id)) {
-            throw new UnauthorizedException();
+            throw new UnauthorizedException;
```
loki-sinclair-hdruk, line 1601: Linter battles again?

app/Providers/AppServiceProvider.php
loki-sinclair-hdruk: And it should have! Using "string" and dot concatenation is weird.

dnhdruk (approved):

app/Http/Controllers/Api/V1/CohortRequestController.php
calmacx: yeh :/

loki-sinclair-hdruk (approved):

PR #1541, opened 2026-02-19
feat(GAT-8541): CRUK API PR

cocoon02:


app/Http/Controllers/Api/V1/AuthController.php
```diff
@@ -79,7 +87,16 @@ public function checkAuthorization(Request $request)
         try {
             $input = $request->all();
 
-            $user = User::where('email', $input['email'])->where('provider', Config::get('constants.provider.service'))->first();
+            $provider = $input['provider'] ?? Config::get('constants.provider.service');
+            $allowedProviders = [
+                Config::get('constants.provider.service'),
```
loki-sinclair-hdruk: This can't be here. It invalidates the comparison from the original `$provider` assignment. Looks like we'd be comparing two identical arrays at all times on the lines below.

Isn't adding `cruk` to the `constants.provider.service` enough?

app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -163,7 +163,7 @@ public function datasets(Search $request): JsonResponse|BinaryFileResponse
 
         try {
             $aggs = Filter::where('type', 'dataset')->where('enabled', 1)->get()->toArray();
-            $input['aggs'] = $aggs;
```
JamieByrneHDRUK, line 166: This will break the gateway search its required here: https://github.com/HDRUK/search-service/blob/dev/pkg/search.go#L68

app/Http/Controllers/Api/V1/AuthController.php
```diff
@@ -170,6 +187,248 @@ public function refreshToken(Request $request)
...
+            $input = $request->validated();
+
+            // Build name from firstname and lastname if provided
+            $name = trim(($input['firstname'] ?? '') . ' ' . ($input['lastname'] ?? ''));
+            if (empty($name)) {
+                $name = $input['email']; // Fallback to email if no name provided
+            }
+
+            // Create user
+            $provider = $input['provider'] ?? Config::get('constants.provider.service');
```
JamieByrneHDRUK: This should only be for CRUK provider not other users.

app/Http/Controllers/Api/V1/AuthController.php
```diff
@@ -170,6 +187,248 @@ public function refreshToken(Request $request)
...
+     *       description="Invalid credentials",
+     *    ),
+     * )
+     */
+    public function login(LoginRequest $request)
+    {
+        try {
+            $input = $request->validated();
+
+            $provider = $input['provider'] ?? Config::get('constants.provider.service');
```
JamieByrneHDRUK: CRUK only please

app/Http/Controllers/Api/V1/FormHydrationController.php
```diff
@@ -127,9 +127,10 @@ public function index(Request $request): JsonResponse
      */
     public function onboardingFormHydration(Request $request): JsonResponse
     {
-        $model = $request->input('model', Config::get('form_hydration.schema.model'));
+        $model = $request->input('name', $request->input('model', Config::get('form_hydration.schema.model')));
```
JamieByrneHDRUK: This should really be being controlled using a middleware, detect the CRUK session header and set the model at app level for the schema model to be the model/version you require.

app/Http/Controllers/Api/V1/AuthController.php
cocoon02: Done.  Thanks for the feedback.

app/Http/Controllers/Api/V1/SearchController.php
cocoon02: Done.

app/Http/Controllers/Api/V1/AuthController.php
cocoon02: Provider from the request is now only used when it’s explicitly CRUK; otherwise we default to service and ignore other values

app/Http/Controllers/Api/V1/AuthController.php
cocoon02: Done

app/Http/Controllers/Api/V1/FormHydrationController.php
cocoon02: Schema model/version are no longer taken from the request; they’re set in middleware when the X-CRUK-Session header is present, and the controller uses config only.

cocoon02:
@JamieByrneHDRUK the ci/cd requires approval as I have updated the PR. Thanks

app/Http/Controllers/Api/V1/FormHydrationController.php
```diff
@@ -19,24 +19,13 @@ class FormHydrationController extends Controller
...
-     *      @OA\Parameter(
-     *          name="version",
-     *          in="query",
-     *          required=false,
-     *          description="The version of the model for which form schema is requested.",
-     *          @OA\Schema(
-     *              type="string"
-     *          )
+     *          description="When present, form hydration uses the CRUK schema (model/version).",
+     *          @OA\Schema(type="string")
```
calmacx: why is this being changed?

app/Http/Controllers/Api/V1/FormHydrationController.php
```diff
@@ -57,8 +46,8 @@ class FormHydrationController extends Controller
      */
     public function index(Request $request): JsonResponse
     {
-        $model = $request->input('model', Config::get('form_hydration.schema.model'));
-        $version = $request->input('version', Config::get('form_hydration.schema.latest_version'));
+        $model = Config::get('form_hydration.schema.model');
+        $version = Config::get('form_hydration.schema.latest_version');
```
calmacx: why change this? This is our critical hydration endpoint

config/jwt.php
```diff
@@ -2,6 +2,6 @@
 
 return [
     "kid" => env("JWT_KID", "jwtkidnotfound"),
-    "secret" => env("JWT_SECRET", 300),
+    "secret" => env("JWT_SECRET", ""),
```
calmacx: ?

database/seeders/omop/ConceptSeeder.php
```diff
@@ -91,4 +89,18 @@ public function run()
...
+            DB::statement('SET FOREIGN_KEY_CHECKS = 0;');
+        }
+    }
+
+    private function enableForeignKeyChecks(): void
+    {
+        if (!app()->environment('testing') && strtolower(DB::connection()->getDriverName()) === 'mysql') {
+            DB::statement('SET FOREIGN_KEY_CHECKS = 1;');
+        }
+    }
```
calmacx: global function if this is gonna be changed - but also dont get why this is being changed

public/frankenphp-worker.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+// Set a default for the application base path and public path if they are missing...
+$_SERVER['APP_BASE_PATH'] = $_ENV['APP_BASE_PATH'] ?? $_SERVER['APP_BASE_PATH'] ?? __DIR__.'/..';
+$_SERVER['APP_PUBLIC_PATH'] = $_ENV['APP_PUBLIC_PATH'] ?? $_SERVER['APP_BASE_PATH'] ?? __DIR__;
+
+require __DIR__.'/../vendor/laravel/octane/bin/frankenphp-worker.php';
```
calmacx, line 7: remove file

app/Http/Controllers/Api/V1/FormHydrationController.php
cocoon02: We changed this to match a review request: schema model/version are no longer taken from the request. They’re set in app config, and when the client sends the X-CRUK-Session header, the CrukSessionHeader middleware applies the schema (same env/config as default: FORM_HYDRATION_SCHEMA_MODEL / FORM_HYDRATION_SCHEMA_LATEST_VERSION). The OpenAPI block was updated so the docs describe this behaviour (header instead of query params).

app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -175,33 +175,43 @@ public function datasets(Search $request): JsonResponse|BinaryFileResponse
...
             $response = $response->json();
 
-            if (
-                !isset($response['hits']) || !is_array($response['hits']) ||
-                !isset($response['hits']['hits']) || !is_array($response['hits']['hits']) ||
-                !isset($response['hits']['total']['value'])
-            ) {
+            // Handle various response formats more robustly
+            $hits = $response['hits'] ?? [];
+            $hitsArray = is_array($hits) ? ($hits['hits'] ?? []) : [];
```
dnhdruk: `$hits` is all time array

app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -175,33 +175,43 @@ public function datasets(Search $request): JsonResponse|BinaryFileResponse
...
+            $datasetsArray = is_array($hitsArray) ? $hitsArray : [];
```
dnhdruk: `$hitsArray` is all time array

database/seeders/omop/ConceptSeeder.php
cocoon02: Because on MySQL, truncating tables that have foreign keys between them fails unless we temporarily disable FK checks. The seeders do that so the truncates can run; we skip it in the testing env (SQLite doesn’t support it).

public/frankenphp-worker.php
cocoon02: Removed

app/Http/Controllers/Api/V1/FormHydrationController.php
cocoon02: We changed this to match a review request: schema model/version are no longer taken from the request. They’re set in app config, and when the client sends the X-CRUK-Session header, the CrukSessionHeader middleware applies the schema (same env/config as default: FORM_HYDRATION_SCHEMA_MODEL / FORM_HYDRATION_SCHEMA_LATEST_VERSION)

app/Http/Controllers/Api/V1/SearchController.php
cocoon02: Updated !

app/Http/Controllers/Api/V1/SearchController.php
cocoon02: updated !

config/jwt.php
cocoon02: Updated

app/Http/Controllers/Api/V1/SearchController.php
```diff
@@ -175,33 +175,43 @@ public function datasets(Search $request): JsonResponse|BinaryFileResponse
...
+            
+            // Handle total in various formats (object with 'value' or direct number)
+            $total = $hits['total'] ?? null;
+            if (is_array($total) && isset($total['value'])) {
+                $totalResults = (int) $total['value'];
+            } elseif (is_numeric($total)) {
+                $totalResults = (int) $total;
+            } else {
+                $totalResults = count($datasetsArray); // Fallback to count
+            }
```
dnhdruk: changes to this method are not necessary

app/Http/Controllers/Api/V1/TeamController.php
```diff
@@ -1558,6 +1618,208 @@ function ($query) {
...
+     * @param Request $request
+     * @param int $teamId
+     * @return JsonResponse
+     */
+    public function datasetsCohortDiscovery(Request $request, int $teamId): JsonResponse
+    {
+        try {
+            $filterStatus = $request->query('status', null);
+            $datasetId = $request->query('dataset_id', null);
+            $mongoPId = $request->query('mongo_pid', null);
```
dnhdruk, line 1663: why?

app/Http/Controllers/Api/V1/TeamController.php
```diff
@@ -1558,6 +1618,208 @@ function ($query) {
...
+                    }
+                )
+                ->when($filterStatus, function ($query) use ($filterStatus) {
+                    return $query->where('status', '=', $filterStatus);
+                })
+                ->when(
+                    $sortOnMetadata,
+                    fn ($query) => $query->orderByMetadata($sortField, $sortDirection),
+                    fn ($query) => $query->orderBy($sortField, $sortDirection)
+                )
```
dnhdruk, line 1758: JSON fields lack indexes, and this statement performs very poorly.

app/Http/Controllers/Api/V1/SearchController.php
cocoon02: Updated.

database/seeders/omop/ConceptSeeder.php
calmacx: hi @cocoon02 - I'll clarify - 
I meant..
1) if this is to be changed, this should be a global function for all seeders not just the ConceptSeeder ideally
2) why is this being touched in the first place - this is specifically a seeder used for a specific (now redundant) service - there's no reason to touch this file - unless you can tell us why?

app/Http/Controllers/Api/V1/FormHydrationController.php
calmacx: Sorry but I cannot resolve this - I believe this will break our onboarding form in our frontend which uses this endpoint:

```
return get<FormHydrationSchema>(
        `${
            apis.formHydrationV1UrlIP
        }?name=${schemaName}&version=${schemaVersion}&dataTypes=${
            dataTypes || []
        }${teamId && `&team_id=${teamId}`}`
    );
```

if you need it work differently for you, can you extend or provide more detail of how this will not break our onboarding form and our existing tests etc. ?

config/jwt.php
calmacx: i do not see this updated - please can you resolve

app/Http/Controllers/Api/V1/FormHydrationController.php
calmacx: This is not resolved as commented below - my comment here was related to why the swagger text has been changed to say "When present, form hydration uses the CRUK schema (model/version)."

config/jwt.php
calmacx: Apologies - i think github might be having some issues and it was not showing me the change reverted

app/Http/Middleware/CrukSessionHeader.php
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @param  \Closure(\Illuminate\Http\Request): (\Symfony\Component\HttpFoundation\Response)  $next
+     */
+    public function handle(Request $request, Closure $next): Response
+    {
+        $value = $request->header(self::CRUK_SESSION_HEADER);
+        if ($value !== null && $value !== '') {
+            $model = Config::get('form_hydration.schema.model', 'HDRUK');
+            $version = Config::get('form_hydration.schema.latest_version', '4.1.0');
+            Config::set('form_hydration.schema.model', $model);
```
loki-sinclair-hdruk: This is not what was suggested. It was suggested that this configuration was set at the application level, not the configuration level, which would change it for _every_ instance of the API. Rather than for the life of the request.

cocoon02: Done.

app/Http/Controllers/Api/V1/TeamController.php
cocoon02: Reverted.

app/Http/Controllers/Api/V1/TeamController.php
cocoon02: Reverted.

database/seeders/omop/ConceptSeeder.php
cocoon02: Some tests were failing. I have reverted.

app/Http/Controllers/Api/V1/FormHydrationController.php
cocoon02: Reverted

app/Http/Controllers/Api/V1/FormHydrationController.php
cocoon02: To avoid breaking the existing onboarding form that calls the endpoint , the API still accepts request params and uses them when present, and only falls back to config when they’re omitted.

PR #1558, opened 2026-02-26
feat(GAT-0000): Optimising auth login from ~3s to < 1s

loki-sinclair-hdruk:
## Describe your changes

Optimises the Gateway authentication path. Details below:

1. `UserController@show` was wasting cycles calling `$user->count()` which runs subsequent queries, whereas `count($user)` does the same without additional db calls. 
2. `JwtMiddleware@handle` was rebuilding roles and perms already eagerly loaded by relations, for every single request. I've refactored this to batch this info.
3. `UserTransformation@getUsers` this entire thing needs refactoring, but for now I've refactored just this method to batch team and role lookups, instead of calling `TeamHasUser` for every single team. Also fixed an N+1 issue whereby `Notifications` were (again) already eagerly loaded, instead of re-querying `user_has_notifications` + `notifications` per user/per notification, which was _really heavy waste_!

*To note* - I've commented on this PR to highlight a specific intentional (but not long term) narrowing. Upon testing the refactor, there were a number of failed tests in completely unrelated areas. This is due to mixed logic requirements for data guarding. Some controllers only require `jwt.verify` middleware, but expect the same roles/perms to be available like in `check.access` - I don't have time to tackle this too, so the narrowing applies these changes just to auth flow.

## Issue ticket link
None. Technically part of DP-504 - and exposed when optimising and adding a loader.

None.

No.

```
composer run phpstan
> vendor/bin/phpstan analyse app  --memory-limit 512M
Note: Using configuration file /Users/lokisinclair/Development/HDRUK/gateway-api-2/phpstan.neon.
 730/730 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
                                                                                                                   
 [OK] No errors                                                                                                                                                             
```

```
  Tests:    600 passed (8896 assertions)
  Duration: 125.12s
  Parallel: 11 processes
```

### DIFF
--- a/app/Http/Middleware/JwtMiddleware.php
+++ b/app/Http/Middleware/JwtMiddleware.php
@@ -59,13 +59,7 @@ class JwtMiddleware
 
             $request->merge(
                 [
-                    'jwt_user' => [
-                        'id' => $user->id,
-                        'name' => $user->name,
-                        'email' => $user->email,
-                        'is_admin' => $user->is_admin,
-                        'role_perms' => $this->getUserRolePerms($user->id),
-                    ],
+                    'jwt_user' => $this->buildJwtUserPayload($request, $user),
                 ],
             );
 
@@ -131,13 +125,7 @@ class JwtMiddleware
 
                 $request->merge(
                     [
-                        'jwt_user' => [
-                            'id' => $user->id,
-                            'name' => $user->name,
-                            'email' => $user->email,
-                            'is_admin' => $user->is_admin,
-                            'role_perms' => $this->getUserRolePerms($user->id),
-                        ],
+                        'jwt_user' => $this->buildJwtUserPayload($request, $user),
                     ],
                 );
                 return $next($request);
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/JwtMiddleware.php
```diff
@@ -152,4 +140,41 @@ private function validateUserId(int $userId)
...
+
+    private function routeRequiresRolePerms(Request $request): bool
+    {
+        $route = $request->route();
+        if (!$route) {
+            return true;
+        }
+
+        $controllerAction = (string) ($route->getAction()['controller'] ?? '');
+        if ($controllerAction === 'App\\Http\\Controllers\\Api\\V1\\UserController@show') {
```
loki-sinclair-hdruk, line 168: Specific narrowing.

tmillingtonhic: Would it be a good idea to leave a comment about this?

dnhdruk (approved):

tmillingtonhic (approved):

calmacx (approved):

app/Http/Middleware/JwtMiddleware.php
loki-sinclair-hdruk: Absolutely there is, but I don't see this being around long-term. There is significant risk in mixing business concerns with middleware and this is a prime example.

PR #1569, opened 2026-03-04
feat(GAT-0000): Refactoring to enable partner context for API operation for CRUK and others, and optimisations when noticed

loki-sinclair-hdruk:
number of optimisations noticed while refactoring. Fixes N+1, and N+1 2N problems. Makes use of existing data to avoid duplicated queries. Eliminates full table ID scan when no title filter. A couple of instances of fetching only needed columns instead of full roes. An instance of 2 queries being replaced by 1.

## Describe your changes

Significant changes for abstracting Dataset functionality away from Controller bloat and into specific Services. First pass for CRUK that enables `x-partner-context` to be sent through as a request header that dictates routes and transformations as needed. Defaults to HDRUK for our own unaffected operation

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-8667

### DIFF
--- a/app/Http/Controllers/Api/V2/DatasetController.php
+++ b/app/Http/Controllers/Api/V2/DatasetController.php
@@ -32,7 +32,8 @@ class DatasetController extends Controller
     public function __construct(
         private readonly DatasetService $datasetService,
         private readonly PartnerContext $partnerContext,
-    ) {}
+    ) {
+    }
 
     /**
      * @OA\Get(
--- a/app/Http/Controllers/Api/V2/DatasetController.php
+++ b/app/Http/Controllers/Api/V2/DatasetController.php
@@ -2,46 +2,37 @@
 
 namespace App\Http\Controllers\Api\V2;
 
-use Config;
 use Auditor;
+use Config;
 use Exception;
-use App\Models\Team;
-use App\Models\User;
 use App\Models\Dataset;
-use App\Jobs\TermExtraction;
-use App\Jobs\LinkageExtraction;
-use Illuminate\Http\Request;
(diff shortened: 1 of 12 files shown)


dnhdruk (approved):

tmillingtonhic (approved):

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -207,94 +159,58 @@ public function index(Request $request): JsonResponse
...
-     *
      */
     public function showActive(GetDataset $request, int $id): JsonResponse|BinaryFileResponse
     {
         try {
-            $exportStructuralMetadata = $request->query('export', null);
-
-            // Retrieve the dataset with collections, publications, and counts
-            $dataset = Dataset::with("team")->where("status", Dataset::STATUS_ACTIVE)->find($id);
+            $dataset = $this->datasetService->findActive($id);
```
spco, line 176: This drops the `with("team")` clause - long time since I touched this, but have you confirmed that's never used from the payload by the FE?

spco (approved):
There are enough changes to not be 100%, but looks good. One comment.

app/Http/Controllers/Api/V2/DatasetController.php
loki-sinclair-hdruk: When I tested the returns from both dev and this branch locally, the payloads were identical, for all routes. Thanks in part due to making use of previously loaded relations and such. But, I will just double check to confirm again, as it was a week ago now.

spco (approved):

PR #1608, opened 2026-04-07
feat(GAT-8692): Download Custodian dashboard information - PDF

dnhdruk:
## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-8692


app/Http/Controllers/Api/V3/TeamDashboardController.php
```diff
@@ -462,4 +465,166 @@ public function downloadCsv(GetTeamDashboard $request, $id)
...
+            $startDate = now()->subYear()->format('Y-m-d');
+            $endDate = now()->format('Y-m-d');
+        }
+
+        $entityDatasets = $this->teamDashboardService->getCount(Dataset::class, 'active_date', $id, $startDate, $endDate, ['status'  => Dataset::STATUS_ACTIVE]);
+        $entityDataUses = $this->teamDashboardService->getCount(Dur::class, 'active_date', $id, $startDate, $endDate, ['status'  => Dur::STATUS_ACTIVE]);
+        $entityTools = $this->teamDashboardService->getCount(Tool::class, 'active_date', $id, $startDate, $endDate, ['status'  => Tool::STATUS_ACTIVE]);
+        $entityPublications = $this->teamDashboardService->getCount(Publication::class, 'active_date', $id, $startDate, $endDate, ['status'  => Publication::STATUS_ACTIVE]);
+        $entityCollections = $this->teamDashboardService->getCount(Collection::class, 'active_date', $id, $startDate, $endDate, ['status'  => Collection::STATUS_ACTIVE]);
+        $entityGeneralEnquiries = $this->teamDashboardService->getCount(EnquiryThread::class, 'created_at', $id, $startDate, $endDate, ['is_general_enquiry' => 1]);
```
calmacx, line 540: lint doesnt split this on separate lines to make more readable?

app/Http/Controllers/Api/V3/TeamDashboardController.php
```diff
@@ -462,4 +465,166 @@ public function downloadCsv(GetTeamDashboard $request, $id)
...
+        $endDate = $request->query('endDate') ?? null;
+
+        if ($startDate && $endDate && $startDate > $endDate) {
+            return $this->errorResponse('startDate must be less than or equal to endDate');
+        }
+
+        if ($startDate === null || $endDate === null) {
+            $startDate = now()->subYear()->format('Y-m-d');
+            $endDate = now()->format('Y-m-d');
+        }
```
calmacx, line 533: what if only one is `null` ? the other will be overwritten?

app/Http/Controllers/Api/V3/TeamDashboardController.php
dnhdruk: no

app/Http/Controllers/Api/V3/TeamDashboardController.php
dnhdruk: If one or both are not received at that time, I will provide default values ​​(one year interval) - is a protection

app/Services/DashboardSvgCharts.php
```diff
@@ -0,0 +1,214 @@
...
+    // ─────────────────────────────────────────────
+
+    private function niceMax(array $values): int
+    {
+        $max = max($values);
+        $exp = (int) floor(log10($max));
+        $pow = 10 ** $exp;
+
+        return (int) (ceil($max / $pow) * $pow);
+    }
```
calmacx, line 213: so BE is producing charts to put into a PDF rather than the FE generating something that can be downloaded?

weird... but ok if is the requirement??

resources/views/dashboard/show.blade.php
```diff
@@ -0,0 +1,306 @@
...
+          <span class="row-label">Data Access Requests</span>
+          <span class="row-num">16</span>
+        </div>
+      </div>
+    </div>
+    </div>
+ 
+</div>
+</body>
+</html>
```
calmacx, line 306: strange this has been a requirement to me

app/Http/Controllers/Api/V3/TeamDashboardController.php
calmacx: very long lines of code, strange it is not split onto separate lines - hard to read

app/Services/DashboardSvgCharts.php
dnhdruk: We have two methods: one is the backend (which is this) and, using the frontend, we have another ticket. It depends on which one will be more ok and the best method will be chosen.

app/Http/Controllers/Api/V3/TeamDashboardController.php
calmacx: is it possible to have one null and the other not null? because in that scenario this protection is used with a random yearly start/end date?

resources/views/dashboard/show.blade.php
dnhdruk: Convert blade to HTML and HTML to PDF on the go.

app/Http/Controllers/Api/V3/TeamDashboardController.php
dnhdruk: Actually it's not too long but maybe because we need several different data to send to create the pdf

app/Services/DashboardSvgCharts.php
loki-sinclair-hdruk: I'm with @calmacx on this @dnhdruk. BE **should never** generate graphics. Only the data for the series' required to fill said graphics. This is flooding the BE with image generation, which is not what we want to be doing.

Where was the plan for this? At what point did someone decide that the API should be creating images? I need to see the ownership chain here, because it's weird.

loki-sinclair-hdruk:
Not at all sure this is the way to go with this. The content is already rendered in the frontend, surely it would be easier to pull a PDF export module into that and pass the rendered components, instead of rebuilding the report for a second time and then exporting from the BE.

app/Services/DashboardSvgCharts.php
dnhdruk: In this case, this task will close with the mention of having completed the pdf export in the frontend and the pr will not be merged.

PR #1636, opened 2026-05-12
fix(GAT-8785): Introduces a few fixes for edge-cases in gmi. Also moves GMI jobs to separate queue to avoid block and drops

loki-sinclair-hdruk:

- [x] I have created mocks for unit tests (where appropriate)
- [x] I have added audit logs for new operation logic (if applicable)

### DIFF
--- a/app/Jobs/ProcessFederation.php
+++ b/app/Jobs/ProcessFederation.php
@@ -58,7 +58,7 @@ class ProcessFederation implements ShouldQueue
         try {
             // Here and not in constructor because this library makes excessive use
             // of closures which can't be serialised by Laravel cache.
-            $this->gsms = new GoogleSecretManagerService();
+            $this->gsms = app(GoogleSecretManagerService::class);
             $remoteItems = $this->pullCatalogueList($this->federation, $this->gsms);
 
             if ($remoteItems->isEmpty()) {
--- a/app/Jobs/ProcessFederation.php
+++ b/app/Jobs/ProcessFederation.php
@@ -36,6 +36,7 @@ class ProcessFederation implements ShouldQueue
 
     public int $tries = 3;
     public int $backoff = 60;
+    public int $timeout = 120;
 
     /**
      * Create a new job instance.
@@ -44,6 +45,7 @@ class ProcessFederation implements ShouldQueue
     {
         $this->federation = $federation;
         $this->gmi = new GatewayMetadataIngestionService();
+        $this->onQueue('federation');
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/Jobs/ProcessFederation.php
```diff
@@ -44,60 +45,65 @@ public function __construct(Federation $federation)
...
-        $this->log('info', "metadata ingestion completed for team {$this->gmi->getTeam()} - created: {$created}, updated: {$updated}, archived: {$archived}");
+            $archived = $this->archiveLocalDatasetsNotInRemoteCatalogue($localItems, $remoteItems, $this->gmi);
 
-        FederationProcessed::dispatch($this->federation);
+            $this->log('info', "metadata ingestion completed for team {$this->gmi->getTeam()} - created: {$created}, updated: {$updated}, archived: {$archived}");
 
-        return;
+            FederationProcessed::dispatch($this->federation);
+        } finally {
+            $this->federation->update(['is_running' => false]);
```
dnhdruk, line 105: I believe you don't need this because it is already addressed by `failed(...)`.

loki-sinclair-hdruk: What about when it's successful?

dnhdruk: `FederationProcessed::dispatch(...)`

dnhdruk (approved):

soringumeni1 (approved):

tests/Feature/ProcessFederationJobTest.php
```diff
@@ -0,0 +1,338 @@
...
+    private function mockGsms(): void
+    {
+        $this->mock(GoogleSecretManagerService::class);
+    }
+
+    private function makeFederation(): array
+    {
+        $team = Team::factory()->create();
+        $federation = Federation::factory()->create([
+            'auth_type' => 'NO_AUTH',
```
calmacx, line 43: any test needed for different auth types? probably not - but just checking

calmacx (approved):

tests/Feature/ProcessFederationJobTest.php
loki-sinclair-hdruk: No right now, no. I think all partners have opted for open catalogs

app/Traits/GatewayMetadataIngestionTrait.php
```diff
@@ -47,14 +47,9 @@ private function getCatalogueFromFederationModel(Federation $federation, GoogleS
...
-        return [
-            'data' => [
-                'errors' => $response->json(),
-                'status' => $response->status(),
-                'success' => false,
-                'title' => 'Test Unsuccessful',
-            ],
-        ];
+        throw new \RuntimeException(
+            "Remote catalogue returned non-200 status {$response->status()} for {$url}: " . $response->body()
```
spco, line 51: What response is the FE "Test-run integration" panel expecting on failure? Will that need updating too?

I'm referring to this guy on the right:

Image

spco (approved):

app/Traits/GatewayMetadataIngestionTrait.php
spco: Yeah after some investigation I think this will break the FE. The `src/hooks/runFederation` will need updating to handle a raised error rather than the specific array format previously returned on error.

On the BE, note that app/Jobs/TestFederation also relies upon receiving the array response format on failure, so would likely want to be updated there

spco (changes requested):

app/Traits/GatewayMetadataIngestionTrait.php
loki-sinclair-hdruk: Damn good catch fella, completely forgot about the test in the FE.

loki-sinclair-hdruk: @spco I'm pretty sure that return would be the same upon closer inspection, as it's passed as an array which makes use of `getCatalogueFromFederationArray` instead - which remains untouched. Will discuss on slack.

spco (approved):

PR #1663, opened 2026-06-10
feat(GAT-8796): allows deferred search providers. allows overriding of provider to search

loki-sinclair-hdruk:
## Describe your changes

Default is all with HDR being the immediate, and others being deferred. Also tinkers with queue priorities to better handle clashes of jobs

Search Aggregation now offers:

* Default - All known providers will run the same criteria for search.
* Allows overriding of provider to search on. Passing `providers` array of what providers to call. HDRUK will always run instantly. External providers will be run as deferred jobs. Polling of endpoint with the cache token provided will allow client to render results when ready.

`SEARCH_DEFERRED_TTL` - allows the overriding of the default (60 second) cache for deferred search jobs

- [x] I have added new environment variables to the .env.example file (if applicable)


app/Http/Controllers/Api/V2/SearchController.php
```diff
@@ -29,16 +31,86 @@ public function search(Search $request): JsonResponse
...
+
+        if (!empty($deferredProviders)) {
+            $token        = 'srch_' . Str::random(32);
+            $ttl          = config('gateway.search_deferred_ttl');
+            $pendingNames = array_map(fn ($p) => $p->getShortName(), $deferredProviders);
+
+            Cache::put("search:deferred:{$token}:pending", $pendingNames, $ttl);
+
+            foreach ($deferredProviders as $provider) {
+                DeferredProviderSearch::dispatch(
```
calmacx, line 52: Do you need to error handle the `dispatch` if any of the context providers fail to dispatch it doesnt break the search controller?

app/Jobs/DeferredProviderSearch.php
```diff
@@ -0,0 +1,67 @@
...
+
+class DeferredProviderSearch implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    public $tries   = 2;
+    public $timeout = 30;
```
calmacx, line 20: are these needed or should be the same as what you have in this new `high` queue with 3 tries?

calmacx (approved):

app/Http/Controllers/Api/V2/SearchController.php
loki-sinclair-hdruk: Nah. That's the beauty of it. We don't require them to succeed in the slightest. So, this will silently fail in the background, but end-users will never know and we don't care.

app/Jobs/DeferredProviderSearch.php
loki-sinclair-hdruk: Nope. These relate directly to the deferred queue. With the `high` queue getting more priority in general.

PR #1664, opened 2026-06-11
fix(GAT-8796): Fixes sorting missing from refactor. Fixes naming of n…

loki-sinclair-hdruk:
…ew feature flag consistently. Fixes empty query not returning results after refactor

> AI usage. Claude Sonnet 4.6 was used to generate unit tests to prove recent changes.

### DIFF
--- a/app/SearchProviders/HDRUK.php
+++ b/app/SearchProviders/HDRUK.php
@@ -72,9 +72,12 @@ class HDRUK implements SearchProvider
         try {
             $filterConfig = self::FILTER_TYPE_MAP[$type] ?? ['type' => $type, 'enabledOnly' => false];
 
-            $input          = $params;
-            $input['query'] = $query;
-            $input['aggs']  = FilterCache::get($filterConfig['type'], $filterConfig['enabledOnly']);
+            $input         = $params;
+            $input['aggs'] = FilterCache::get($filterConfig['type'], $filterConfig['enabledOnly']);
+
+            if ($query !== '') {
+                $input['query'] = $query;
+            }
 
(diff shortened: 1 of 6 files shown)


spco (approved):

app/SearchProviders/HDRUK.php
```diff
@@ -117,6 +121,38 @@ public function search(string $query, string $type, array $params = []): array
...
+
+                if (is_string($aVal) && strtotime($aVal) !== false) {
+                    return strtotime((string)$bVal) <=> strtotime((string)$aVal);
+                } elseif (is_string($aVal)) {
+                    return strtoupper((string)$bVal) <=> strtoupper((string)$aVal);
+                } else {
+                    return $bVal <=> $aVal;
+                }
+            });
+        }
```
calmacx: how come the `else` statement here (I captures when `$direction` is set to be `desc`) is doing this `strtotime` when the `asc` is not (it is just doing ` $a['_source'][$field] <=> $b['_source'][$field]` ?

calmacx (approved):
approved with a question for clarification

app/SearchProviders/HDRUK.php
loki-sinclair-hdruk: Good spot! I just took this as gospel from the old controller logic, that I missed in the refactoring to the hydrators. Looking at it, it looks to be a fluke that this was working, especially with dates - likely because we use Carbon. Let me tinker.

loki-sinclair-hdruk: @calmacx Updated. Also had Claude write up some new tests to prove.

PR #1666, opened 2026-06-16
feat(GAT-0000): seed 20 CRUK datasets from cruk_dummy_data fixtures

cocoon02:
Replace generic random dataset versions with one-to-one CRUK metadata seeding so local dev matches partner_context isolation and project grant demo data.

### DIFF
--- a/database/seeders/DatasetSeeder.php
+++ b/database/seeders/DatasetSeeder.php
@@ -12,6 +12,8 @@ class DatasetSeeder extends Seeder
      */
     public function run(): void
     {
-        Dataset::factory()->count(10)->create();
+        Dataset::factory()->count(20)->create([
+            'partner_context' => 'CRUK',
+        ]);
     }
 }
--- a/database/seeders/CRUKDatasetSeeder.php
+++ b/database/seeders/CRUKDatasetSeeder.php
@@ -5,13 +5,15 @@ namespace Database\Seeders;
 use App\Models\Dataset;
 use Illuminate\Database\Seeder;
 
-class DatasetSeeder extends Seeder
+class CRUKDatasetSeeder extends Seeder
 {
     /**
      * Run the database seeds.
      */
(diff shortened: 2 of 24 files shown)


database/seeders/DatasetSeeder.php
```diff
@@ -12,6 +12,8 @@ class DatasetSeeder extends Seeder
      */
     public function run(): void
     {
-        Dataset::factory()->count(10)->create();
+        Dataset::factory()->count(20)->create([
```
spco: We wouldn't want you editing this Seeder directly with this context, otherwise when we run it it will mess our data up. Instead, please create a `CRUKDatasetSeeder` with this in it. Same for DatasetVersion, make a `CRUKDatasetVersionSeeder`.

You can then probably simply run something like
```
php artisan db:seed --fresh
php artisan db:seed --class=CRUKDatasetSeeder
php artisan db:seed --class=CRUKDatasetVersionSeeder
```
to give you datasets that belong to both `hdruk` and `cruk` partner contexts. Or you could do something fancier if you _only_ want to seed CRUK style datasets by making the partner context be read in as appropriate.

spco (changes requested):
See comment. We need to ensure that the "default" behaviour of seeders remains compatible with HDRUK rather than CRUK.

database/seeders/DatasetSeeder.php
cocoon02: Done — thanks for the feedback.

spco (approved):
Looks good.

Before I merge, can I just clarify the expectation here? Once merged, I would run these 2 new seeders in the dev environment to create dummy datasets in the CRUK partner context. You can then test pointing your FE at our dev api and hopefully that will then show only those CRUK datasets.

Is my understanding correct?

cocoon02:
@spco That's correct.

PR #1667, opened 2026-06-16
feat(GAT-9012): First round of template work - refactor the seeder fi…

loki-sinclair-hdruk:
…le as it was getting out of control.

> Claude Sonnet 4.6 was used extensively during this refactor. It was instructed to abstract individual templates from the EmailSeeder file and create new json chunks for isolated template files, by identifier name. 

## Screenshots (if relevant)
image

## Describe your changes
Adds in two new templates per the ticket requirements. No business logic change to use said templates yet. That will come next.

Refactors EmailTemplateSeeder to glob a new `database/seeders/email-templates` directory, which houses all templates as isolated .json files, to mean easier maintenance in the future. Updates the seeder to import directly from these files and seed in a loop. 

Adds in a preview email template command to enable quick visibility of mjml templates. Screenshot attached. Preview allows overriding of layout used `--layout=<new|standard>`. All standing templates use standard as default. The two new ones use new, per the new design preference.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-9012

Not yet


spco (commented):
My main concern is that you're hardcoding the header etc each time individually. Wouldn't it be preferable to maintain the structure of using `$mjmlHead` etc from the existing EmailTemplateSeeder? Even if we then use the `replacement` structure to inject them. DRY and all that. I hated having to modify 70 templates when we modified something in the header.

loki-sinclair-hdruk:
> My main concern is that you're hardcoding the header etc each time individually. Wouldn't it be preferable to maintain the structure of using `$mjmlHead` etc from the existing EmailTemplateSeeder? Even if we then use the `replacement` structure to inject them. DRY and all that. I hated having to modify 70 templates when we modified something in the header.

This is a good shout. In theory the header and/or footer shouldn't be changing now, but that's not very future proof. Let me have a think.

spco (approved):
Looks fine to me.

Two questions: in the new design, are we ditching the copyright text? And it looks like in the new template there's no link associated to the image - should that click through to the Gateway homepage? Since the "standard" template included a link in the footer and otherwise there's no link in the email.

If you can add in the subject line into the previews (maybe in the tab title?) then you'll get an extra sticker for your chart ⭐ but it's not crucial

spco (approved):
All looks sensible

PR #1674, opened 2026-06-22
feat(GAT-8501): Make TeamAdmin an actual admin

philreekshdr:
## Describe your changes
Make TeamAdmin user an actual admin so the feature flags can be toggled in cypress tests

This is a seeder that is run on each cypress test run

### DIFF
--- a/database/seeders/TestDevUsersSeeder.php
+++ b/database/seeders/TestDevUsersSeeder.php
@@ -18,7 +18,7 @@ class TestDevUsersSeeder extends Seeder
         $teamId = 21;
 
         $users = [
-             ['name' => 'TeamAdmin', 'role_ids' => [7], 'email' => 'CypressTeamAdmin@hdruk.ac.uk'],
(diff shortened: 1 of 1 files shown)


spco (commented):
Is this really what you want? `User::is_admin` denotes a super-user I think, not a team admin.

philreekshdr:
> Is this really what you want? `User::is_admin` denotes a super-user I think, not a team admin.

Yep, i need a way of enabling and disabling the feature flags in the cypress tests. This will allow me to control them using the token and hitting the feature flag endpoint

spco (commented):
Right. I can see the aim. I'm not sure this is the perfect way to do this, as this will affect all other uses of `TeamAdmin` elsewhere in cypress tests, and that can have a wide-ranging effect on what's shown. Could you perhaps use a dedicated user rather than piggy-back on `TeamAdmin`? That would avoid any unexpected surprises later on when we forget that `TeamAdmin` happens to be a superuser.

philreekshdr:
Updated to use a name of superuser

spco (approved):

PR #1695, opened 2026-07-07
feat(GAT-1111): add ProjectGrant create endpoint

cocoon02:
**Summary**

Adds POST /api/v1/project_grants to create a ProjectGrant plus initial version.
Adds request validation via CreateProjectGrant.
Adds feature coverage for creating a grant (incl. version + dataset link).

**Test plan**
php artisan test --filter ProjectGrantTest


app/Http/Controllers/Api/V1/ProjectGrantController.php
```diff
@@ -192,6 +195,89 @@ public function show(Request $request, int $id): JsonResponse
...
+     *    )
+     * )
+     */
+    public function store(CreateProjectGrant $request): JsonResponse
+    {
+        try {
+            $input = $request->validated();
+
+            $grant = ProjectGrant::create([
+                'pid' => $input['pid'],
```
spco, line 232: If these are meant to be unique, might it be worth checking?

config/routes.php
```diff
@@ -3486,6 +3486,15 @@
         'middleware' => [],
         'constraint' => [],
     ],
+    [
+        'name' => 'project_grants.store',
+        'method' => 'post',
+        'path' => '/project_grants',
+        'methodController' => 'ProjectGrantController@store',
+        'namespaceController' => 'App\Http\Controllers\Api\V1',
+        'middleware' => [],
```
spco: Must add the correct middleware. At least `'jwt.verify'` but we probably also want to add in specific permissions for `projectgrant.create` etc into the permissions model. That is likely a bigger piece of work though,

app/Http/Controllers/Api/V1/ProjectGrantController.php
```diff
@@ -192,6 +195,89 @@ public function show(Request $request, int $id): JsonResponse
...
+     *          @OA\Property(property="data", type="object")
+     *        )
+     *    )
+     * )
+     */
+    public function store(CreateProjectGrant $request): JsonResponse
+    {
+        try {
+            $input = $request->validated();
+
```
spco, line 230: Please ensure you use the `getAccessorUserAndTeam` trait to standardise and validate the input payload, and then use `checkAccess` to control access.

TBH we need to agree the ownership model for ProjectGrants - are they owned by Teams, by Users, or by either? See also later comment about permissions model needing to be extended in light of that.

app/Http/Controllers/Api/V1/ProjectGrantController.php
loki-sinclair-hdruk: I too question this. For object creation I'd imagine that this would be generated in real-time via UUID or something similar, but this appears to come from the client, is that correct? Does the client create unique pids?

tests/Feature/ProjectGrantTest.php
```diff
@@ -133,4 +133,52 @@ public function test_show_returns_404_for_unknown_id(): void
             'data' => 'Not Found',
         ]);
     }
+
+    public function test_store_creates_project_grant_with_version_and_dataset_links(): void
```
loki-sinclair-hdruk, line 151: Once the middleware for `jwt.verify` has been included above, per Sam's comment, this test will fail and will need addressing.

app/Http/Controllers/Api/V1/ProjectGrantController.php
cocoon02: Looking at the dataset metadata structure (e.g. dataset_00.json), a project grant is supplied as an object where pid is the grant ID:

"projectGrants": [
  {
    "pid": "ORT105",
    "projectGrantName": "Pancreatic Cancer Radiomics Integration",
    "leadResearcher": "Dr. Miller",
    ...
  }
]
So pid (e.g. "ORT105") is the grant identifier and is part of the project grant object the client supplies on create — not something we generate separately. Sarah supplied something like CRCXX-XXXXXXXXX.

app/Http/Controllers/Api/V1/ProjectGrantController.php
cocoon02: Fixed. They are owned by the Teams .

tests/Feature/ProjectGrantTest.php
cocoon02: Fixed.

config/routes.php
cocoon02: Fixed.

app/Http/Controllers/Api/V1/ProjectGrantController.php
cocoon02: Fixed.

loki-sinclair-hdruk (approved):

PR #1697, opened 2026-07-13
feat(GAT-9143): Implements the first migration tests to typesense migration. Datasets and Tools.

loki-sinclair-hdruk:
## Describe your changes
Begins the migration to Typesense. Currently only:

1. Datasets
2. Tools

Governed by two individual feature flags. First being:

 1. V2_SearchAggregation
 2. TypesenseSearch

The first defines whether or not the Gateway uses the new Search Provider layer to funnel off searches. The second determines the search method to use, between Elastic (the fallback) and Typesense.

See .env.example

No.

- [x] I have created mocks for unit tests (where appropriate)
- [x] I have added new environment variables to the .env.example file (if applicable)


app/Models/DatasetVersion.php
```diff
@@ -279,14 +279,38 @@ public function shouldBeSearchable(): bool
...
+            return null;
+        }
+
+        $types = array_values(array_filter(array_map(
+            fn ($item) => ($item['materialType'] ?? null),
+            $tissues
+        ), fn ($type) => $type !== null && $type !== 'None/not available'));
+
+        return $types === [] ? null : array_values(array_unique($types));
+    }
```
calmacx, line 326: You might want to hold off on more changes like this until i'm done with my changes to the GWDM management system.

This is very specific to GWDM 2.0 as your comment says with a hard-coded path

app/Models/DatasetVersion.php
```diff
@@ -308,10 +332,15 @@ public function toSearchableArray(): array
                 data_get($meta, 'metadata.summary.publisher.publisherName', '')
             ),
             'dataType'                      => array_values(array_filter(explode(';,;', $dataType))),
+            'dataSubType'                   => array_values(array_filter(explode(';,;', $dataSubType))),
             'populationSize'                => (int) data_get($meta, 'metadata.summary.populationSize', -1),
             'geographicLocation'            => $this->spatialCoverage->pluck('region')->all(),
             'datasetDOI'                    => data_get($meta, 'metadata.summary.doiName', ''),
             'conformsTo'                    => array_values(array_filter(explode(';,;', $conformsTo))),
+            'accessService'                 => data_get($meta, 'metadata.accessibility.access.accessServiceCategory', '') ?? '',
```
calmacx, line 366: As in my other comment - shall we get in these changes and then put a freeze on it until mine are done? 

My changes are going to have an impact on `data_get` --> this will all be re-written and will need to be context aware (aware of the GWDM version)

calmacx (approved):

app/Models/DatasetVersion.php
calmacx: as per my comment above, this would be better suited in the new GwdmHandler to retrieve materialTypes

app/Models/Team.php
```diff
@@ -350,4 +351,189 @@ public function datasets(): HasMany
...
+            ->filter(fn ($title) => is_string($title) && $title !== '')
+            ->unique()
+            ->values()
+            ->all();
+    }
+
+    private function facetDataTypes(): array
+    {
+        return $this->activeDatasetVersions()
+            ->flatMap(fn ($version) => explode(';,;', data_get($version->metadata, 'metadata.summary.datasetType', '') ?? ''))
```
calmacx, line 413: very specific for GWDM - would break if this changes - will be better in the handlers

app/Http/Controllers/Api/V1/AdminSearchController.php
```diff
@@ -0,0 +1,188 @@
...
+     *         )
+     *     ),
+     *     @OA\Response(response=202, description="Reindex queued"),
+     *     @OA\Response(response=422, description="Unknown entity")
+     * )
+     */
+    public function reindex(Request $request): JsonResponse
+    {
+        try {
+            $validator = Validator::make($request->all(), [
```
equinoxmatt, line 94: Would be nice if we validated this in a custom request object, to avoid the validation logic being in the controller.

app/Jobs/ReindexTypesenseEntity.php
```diff
@@ -0,0 +1,74 @@
...
+            // relations) through toSearchableArray() in chunks, which for
+            // heavier entities (e.g. DatasetVersion's JSON metadata blobs) can
+            // exceed a default CLI memory_limit long before hitting any table
+            // large enough to warrant a permanent php.ini change.
+            ini_set('memory_limit', config('jobs.reindex_memory_limit', '512M'));
+
+            $modelClass = HDRUK::typesenseModelMap()[$this->entity] ?? null;
+
+            if ($modelClass === null) {
+                throw new Exception("Unknown Typesense entity '{$this->entity}'");
```
equinoxmatt, line 52: The failure is deterministic. If it doesn't work now, it won't work on a retry.

Consider using:

```
$this->fail(new Exception("Unknown Typesense entity '{$this->entity}'"));
return;
```

app/Jobs/ReindexTypesenseEntity.php
```diff
@@ -0,0 +1,74 @@
...
+
+            $collectionName = (new $modelClass())->searchableAs();
+
+            if ($typesense->collectionExists($collectionName)) {
+                $typesense->dropCollection($collectionName);
+            }
+
+            $typesense->createCollectionFromModel($modelClass);
+
+            Artisan::call('scout:import', ['model' => $modelClass]);
```
equinoxmatt, line 63: Suggest to use `$modelClass::makeAllSearchable();` - same effect as scout:import and no need to invoke the CLI.

equinoxmatt (approved):

app/Http/Controllers/Api/V1/AdminSearchController.php
loki-sinclair-hdruk: Will discuss offline for pattern to use going forward, as per slack.

PR #1699, opened 2026-07-14
feat(GAT-9212): GWDM version context + handler hierarchy + DatasetService Rearrange

calmacx:
## Screenshots (if relevant)

N/A.

## Describe your changes

Introduces a proper version-handler abstraction for GWDM metadata instead of scattering `version_compare()` calls across `DatasetService`. All version-specific behaviour now goes through `GwdmHandlerFactory`, which resolves a GWDM version string to a handler (`Gwdm1xHandler`, `Gwdm20Handler`, `Gwdm21Handler`, falling back to `Gwdm2xHandler`). Only GWDM 2.x is enabled here — 3.0 support comes later.

- Added `GwdmVersionContext` to resolve the GWDM version per-request from an optional `x-gwdm-version` header (falls back to the configured default on write; returns null on read if not set, so the caller uses the dataset's actual latest version). Also added `OutputSchemaContext` and `PartnerContext` for the same kind of per-request resolution, plus a small `ResolveGwdmVersionContext` middleware to stash the resolved version onto the request early.
- `DatasetService` now delegates to the handler hierarchy instead of branching on version itself, and `V2/DatasetController`, `V1/FormHydrationController`, and `MetadataManagementController` were updated to match.
- Linkage data (dataset-to-dataset and publication-to-dataset relationships) now reads from the `dataset_version_has_dataset_version` / `publication_has_dataset_version` junction tables instead of the raw metadata column, via `Gwdm2xHandler::extractLinkages()` and `DatasetService::getLinkages()`. This also fixes a bug where a delta version row could wipe out linkage data because it doesn't carry full metadata.

Also fixed along the way:
- Version-row insert and its follow-up writes now happen in a single transaction, so a failure partway through can't leave an orphaned version.
- V3's `update()` now dispatches the linkage/term extraction jobs, which it wasn't doing before.
- Changing GWDM schema version between updates now forces a full metadata snapshot instead of trying to diff across incompatible schemas.

This PR should be behaviour-preserving for existing 2.x datasets — nothing here changes what a 2.x read or write actually does, just how the code gets there.

## Issue ticket link
[GAT-9212](https://hdruk.atlassian.net/browse/GAT-9212)

New optional `x-gwdm-version` request header. Added `config/partners.php`.

Yes — adds a `gwdm_version` column to `dataset_versions` (`database/migrations/2026_06_17_000001_add_gwdm_version_to_dataset_versions_table.php`), plus a deployment-step script to backfill it for existing rows.

- New: `GwdmVersionContextTest` (6), `GwdmHandlerFactoryTest` (4), `DatasetVersionSchemaChangeTest` (1), `DatasetVersionGwdmVersionColumnTest` (3).
- Regression: full dataset suite green (`DatasetTest`, `DatasetVersionTest`, `DatasetIntegrationTest` (+Deprecated), `DatasetExportTest`, `AdminDatasetControllerTest`, `DatasetsV2HelpersTest`).
- `composer run phpstan` — clean.

- [x] I have created mocks for unit tests (where appropriate) — TRASER/MMC via `MockExternalApis`
- [ ] I have added appropriate Behat tests to confirm AC (if applicable) — N/A
- [ ] I have added Swagger annotations for new endpoints (if applicable) — no new endpoints
- [ ] I have added audit logs for new operation logic (if applicable) — N/A
- [ ] I have added new environment variables to the .env.example file (if applicable) — N/A
- [ ] I have added new environment variables to terraform repository (if applicable) — N/A

[GAT-9212]: https://hdruk.atlassian.net/browse/GAT-9212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/Models/DatasetVersionHasDatasetVersion.php
```diff
@@ -24,6 +28,9 @@ class DatasetVersionHasDatasetVersion extends Model
         'linkage_type',
         'direct_linkage',
         'description',
+        'raw_url',
+        'raw_pid',
+        'raw_title',
```
calmacx: This is because theres nothing to stop people pushing linkages to external sources (at the moment) - its important to save what was attempted rather than just remove if not match can be found

app/Jobs/LinkageExtraction.php
```diff
@@ -30,69 +27,39 @@ class LinkageExtraction implements ShouldQueue
...
-                return; // Not the correct version for processing
-            }
-
-            // Process dataset and publication linkages
-            $this->processDatasetLinkages();
-            $this->processPublicationLinkages($this->publicationAboutDatasetLinkages, 'ABOUT');
-            $this->processPublicationLinkages($this->publicationUsingDatasetLinkages, 'USING');
-
+            $version = DatasetVersion::findOrFail($this->sourceDatasetVersionId);
+            $handlerFactory->resolve($this->gwdmVersion)->extractLinkages($version);
```
calmacx, line 68: All this functionality in this file hasnt been deleted but has been moved to the GwdmHandlers instead

database/migrations/2026_06_18_000001_add_unresolved_linkage_columns.php
```diff
@@ -0,0 +1,98 @@
...
+return new class extends Migration
+{
+    public function up(): void
+    {
+        $isSqlite = DB::connection()->getDriverName() === 'sqlite';
+
+        Schema::table('dataset_version_has_dataset_version', function (Blueprint $table) use ($isSqlite) {
+            // SQLite rebuilds the whole table on ALTER — named FK drop is not supported.
+            if (! $isSqlite) {
+                $table->dropForeign('ld_dataset_version_target_id_fk');
```
calmacx: because we are making the `dataset_version_target_id` nullable - we are doing this because linkages _could_ be added that fail to match (e.g. external source), so we need to keep the raw values saved and dont have data loss

app/Context/GwdmVersionContext.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    public function targetVersion(): string
+    {
+        $header = request()->header('x-gwdm-version');
+
+        if ($header && in_array($header, GwdmHandlerFactory::supportedVersions(), true)) {
+            return $header;
+        }
+
+        return Config::get('metadata.GWDM.version', '2.0');
```
equinoxmatt, line 41: `targetVersion()` swallows an invalid/unsupported header and falls back to the config default, but `requestedVersion()` right below throws on the same case. Intentional (back-compat?), or should this throw too?

EDIT: I see the test file mentions this, but doesn't state the reason, more that it was intentional.

app/Context/OutputSchemaContext.php
```diff
@@ -0,0 +1,39 @@
...
+ *   2. x-schema-model / x-schema-version request headers — this class
+ *   3. Partner default from config/partners.php           — derived from PartnerContext
+ *   4. null (no translation — return stored GWDM)
+ *
+ * Not needed in queued jobs (output translation is a synchronous read concern).
+ * No middleware or container binding required — instantiated via auto-resolution.
+ */
+class OutputSchemaContext
+{
+    public function __construct(private PartnerContext $partnerContext)
```
equinoxmatt: Can we add `readonly` please?

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -184,10 +186,19 @@ public function showActive(GetDataset $request, int $id): JsonResponse|BinaryFil
                 return $this->streamStructuralMetadataExport($dataset, $id);
             }
 
+            if ($request->query('view') === 'metadataOnly') {
```
equinoxmatt: `view=metadataOnly` is missing from the @OA\Parameter list. Also, is the response shape different here and does that need documenting? The envelope is the same, but not the content.

app/Http/Controllers/Api/V2/DatasetController.php
```diff
@@ -206,6 +217,8 @@ public function showActive(GetDataset $request, int $id): JsonResponse|BinaryFil
 
         } catch (\InvalidArgumentException $e) {
             return response()->json(['message' => $e->getMessage()], 400);
+        } catch (\Illuminate\Database\Eloquent\ModelNotFoundException $e) {
```
equinoxmatt: Proper nitpick, but can we import this at the top, for consistency with the other catch statements.

app/Http/Controllers/Api/V2/DatasetController.php
calmacx: This one crept in from my testing/debugging - I think i'll remove it for a separate PR and/or separate endpoint - thanks

app/MetadataManagementController/MetadataManagementController.php
```diff
@@ -157,13 +157,53 @@ public function validateDataModelType(string &$dataset, string $input_schema, st
...
         }
     }
 
+    /**
+     * Validate a metadata JSON string against the given schema and version, returning
+     * structured error details from TRASER rather than a plain boolean.
+     *
+     * @return array{valid: bool, errors: mixed}
+     */
+    public function validateDataModelTypeWithErrors(string &$dataset, string $input_schema, string $input_version): array
```
equinoxmatt: `$dataset` is never mutated in this method, it's only read (passed into Http::withBody). PHP strings are copy-on-write, so passing by value here is already free, the `&` doesn't optimise anything, it just implies a mutation may be happening when it doesn't.

app/MetadataManagementController/MetadataManagementController.php
```diff
@@ -157,13 +157,53 @@ public function validateDataModelType(string &$dataset, string $input_schema, st
...
+            $urlString = sprintf(
+                '%s/validate?input_schema=%s&input_version=%s',
+                config('services.traser.url'),
+                $input_schema,
+                $input_version
+            );
+
+            $response = Http::withBody($dataset, 'application/json')->post($urlString);
+
+            $valid = $response->status() === 200;
```
equinoxmatt: `isValid` - better naming for bools

app/MetadataManagementController/MetadataManagementController.php
calmacx: Yeah good point - as above, I'm going to break this out into a separate PR and will correct there, since it also crept in as it was originally part of my debugging (validate with errors)

app/MetadataManagementController/MetadataManagementController.php
```diff
@@ -157,13 +157,53 @@ public function validateDataModelType(string &$dataset, string $input_schema, st
...
+
+            return [
+                'valid'  => $valid,
+                'errors' => $valid ? null : $response->json(),
+            ];
+        } catch (Exception $e) {
+            \Log::info($e->getMessage(), $loggingContext);
+            throw new MMCException($e->getMessage());
```
equinoxmatt: `info` feels like the wrong level for an exception?

Also, this is probably something we need to look at another time, but usually when you throw a new exception, you want to pass in the prior exception, which is incredibly useful for debugging:

`throw new MMCException($e->getMessage(), $e);`

In this case I don't think it is possible in the case of MMCException but we should look into it I think.

calmacx: Feel feel to raise a maintenance ticket about this? because I have just copy and pasted how the catch-logging-raising is being handled in this controller. If i was to fix here, I'd have to fix it in many other places (perhaps argument for some trait/middleware to log+raise)

app/Models/Dataset.php
```diff
@@ -202,30 +219,32 @@ function ($query, $fields) {
...
         }
 
-        return null;
+        $result = DB::select(
+            'SELECT dv.id
+             FROM dataset_versions dv
+             WHERE dv.dataset_id = :dataset_id
+             ORDER BY version DESC
+             LIMIT 1',
+            ['dataset_id' => $datasetId]
```
equinoxmatt, line 246: I know this already existed, but the raw SQL caught my eye and I understand why it is being used here. However, correct me if I am wrong, but Datasets and Dataset Versions have soft deletes and this raw sql doesn't seem to filter out soft deleted dataset versions, meaning if a dataset version was deleted it would incorrectly be counted?

app/Context/GwdmVersionContext.php
calmacx: Ive added a comment or two on this but yeh it was intentional because  targetVersion() can't afford to fail because it's called in every request as a context and might not be defined - it allows us to play with >2.1 model versions that are not yet in GwdmHandlerFactory::supportedVersions() or legacy ones.

requestedVersion() can afford to fail loudly because it's only invoked when someone specifically asked a version-related question and can handle a version-related answer.

This is why they are different

app/Models/Dataset.php
calmacx: Good spot. I think you're right here. Needs a check on deleted_at NULL

calmacx: added ` AND dv.deleted_at IS NULL` to both

equinoxmatt (approved):
Thanks for resolving those, looks great.

calmacx:
Great catches @equinoxmatt - gonna have a few more fixes/patches on top too

app/Services/Gwdm/Gwdm1xHandler.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * Keeps the publisher from the incoming metadata payload when its publisherId
+     * maps to an existing team (validated by casting to an integer team id and
+     * looking it up); otherwise falls back to the requesting team.
+     *
+     * KNOWN ISSUE: publisherId here is the raw team primary key, not the team's
+     * persistent id (pid). Properly resolved/normalised on
+     * fix/GAT-9018-publisher-fix.
+     */
+    public function buildPublisher(Team $team, array $incoming = []): array
```
loki-sinclair-hdruk, line 32: As these set the core contract of *MetadataHandler's I'd opt to abstract these signatures into a interface under `app/Contracts` - if you look at `SearchProvider` - you'll see how you can tightly govern these classes from a central location. The service signatures would then be:

```php
class Gwdm1xHandler extends GwdmMetadataHandler implements MetadataHandler
```
...or some other similar naming. It just ensures that each future handler completes the expected contract of methods. Also helps devs down the line when this code invariably becomes buried, as phpstan will throw a fit if anything is missing, or type safety isn't met.

app/Services/Gwdm/Gwdm2xHandler.php
```diff
@@ -0,0 +1,330 @@
...
+ * Differences from 1.x:
+ *   - summary.publisher uses the NEW format: { gatewayId, name }
+ *   - required block includes a 'version' field ("X.0.0")
+ *
+ * 2.0 and 2.1 share the same JSON blob storage strategy — both use
+ * the metadata column with RFC 6902 delta compression. Schema-level
+ * differences (new/changed GWDM fields in 2.1) are handled by TRASER,
+ * not by this handler.
+ */
+class Gwdm2xHandler extends GwdmMetadataHandler
```
loki-sinclair-hdruk, line 26: Same here, obvs.

app/Services/Gwdm/Gwdm2xHandler.php
```diff
@@ -0,0 +1,330 @@
...
+     * Reconstruct the GWDM `linkage` section entirely from SQL junction tables,
+     * making SQL the single source of truth for 2.x linkage data on reads.
+     *
+     * Returns [] (falls through to the stored JSON blob) when no extracted rows
+     * exist — this covers legacy rows that pre-date this migration. Re-dispatching
+     * LinkageExtraction for those versions will backfill the SQL rows.
+     */
+    public function afterRead(DatasetVersion $dv): array
+    {
+        $resolvedDatasets = DatasetVersionHasDatasetVersion::query()
```
loki-sinclair-hdruk: Is this one profiling okay speed wise? `Model::query()` still invokes the eloquent builder, and with a couple of joins, this may be mores responsive as raw sql instead.

app/Services/Gwdm/Gwdm2xHandler.php
```diff
@@ -0,0 +1,330 @@
...
+                'dv.short_title',
+                'd.pid',
+                'd.id as dataset_id',
+            )
+            ->get();
+
+        $publications = PublicationHasDatasetVersion::query()
+            ->where('publication_has_dataset_version.dataset_version_id', $dv->id)
+            ->where('publication_has_dataset_version.description', self::LINKAGE_DESCRIPTION)
+            ->join('publications', 'publications.id', '=', 'publication_has_dataset_version.publication_id')
```
loki-sinclair-hdruk: Same deal. Happy to be told it's performant as-is :)

app/Services/Gwdm/Gwdm1xHandler.php
loki-sinclair-hdruk: Ignore, saw abstracts within the service.

app/Services/Gwdm/Gwdm2xHandler.php
loki-sinclair-hdruk: Ignore.

app/Services/Gwdm/Gwdm2xHandler.php
calmacx: While this was code copied/moved from another place, speed tests didnt show much improvement - switching to SQL as its nicer to read regardless.

app/Services/Gwdm/Gwdm2xHandler.php
calmacx: Switching to SQL as it easier to read and marginally more performant

app/Services/Gwdm/Gwdm2xHandler.php
```diff
@@ -0,0 +1,336 @@
...
+            FROM dataset_version_has_dataset_version
+            INNER JOIN dataset_versions AS dv ON dv.id = dataset_version_has_dataset_version.dataset_version_target_id
+            INNER JOIN datasets AS d ON d.id = dv.dataset_id
+            WHERE dataset_version_has_dataset_version.dataset_version_source_id = ?
+              AND dataset_version_has_dataset_version.direct_linkage = ?
+              AND dataset_version_has_dataset_version.description = ?',
+            [$dv->id, 1, self::LINKAGE_DESCRIPTION]
+        ));
+
+        $publications = collect(DB::select(
```
equinoxmatt: Just a note, turning the results into collections isn't really buying much. The `isNotEmpty()` calls could be good old `!empty()`. 

I would feel differently if the other benefits of being in a collection were being used (mapping, plucking, grouping) etc.

calmacx: yeah good point - reverting to something more simple. Though ive added select on specific columns because if move to raw SQL I lose the type checking on `$row->paper_doi` you get for free with the ORM models (not that that was working when i just used collect())

calmacx:
@loki-sinclair-hdruk @equinoxmatt ive finished with regression and speed testing and this marginally improves speed (locally, with prod db copy) - ive tested it on 600+ datasets - if you can give you final blessings

image
^^ gonna hopefully release this a part of a new tooling/TRASER update in the future

equinoxmatt (approved):

loki-sinclair-hdruk (approved):

loki-sinclair-hdruk:
> @loki-sinclair-hdruk @equinoxmatt ive finished with regression and speed testing and this marginally improves speed (locally, with prod db copy) - ive tested it on 600+ datasets - if you can give you final blessings
> 
> image ^^ gonna hopefully release this a part of a new tooling/TRASER update in the future

Fancy! Homebrew load testing? :D

PR #1706, opened 2026-07-21
feat(GAT-9230): record error and disable after final job failure

equinoxmatt:
ProcessFederation retries a failing job up to three times, before giving up and firing FederationProcessingFailed. The listener for that event only cleared is_running, queued a failure email, and logged the error. The error and error_text columns already present on the federations table were never populated, and a federation that failed kept getting rescheduled indefinitely.

Wire those columns up. On final failure, record the exception message (truncated to the column's 200-char limit) and clear enabled, the same flag getActiveFederations() checks before scheduling a run, so a broken federation stops being retried until someone investigates. ProcessFederationSuccess now clears error/error_text on success, so a federation that is manually re-enabled and starts working again doesn't keep showing a stale error from before.

Add fillable/cast entries for error/error_text on the Federation model, since they were never mass-assignable.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-9230

### DIFF
--- a/app/Listeners/ProcessFederationFailure.php
+++ b/app/Listeners/ProcessFederationFailure.php
@@ -15,7 +15,12 @@ class ProcessFederationFailure
         $federationId = $event->federation->id;
         $jobUuid      = $event->jobUuid;
 
-        $event->federation->update(['is_running' => 0]);
+        $event->federation->update([
+            'is_running' => 0,
+            'error' => true,
+            'error_text' => substr($event->exception->getMessage(), 0, 200),
+            'enabled' => false,
+        ]);
 
         SendEmailCustomIntegration::dispatch($federationId, $jobUuid, 'failure', $event->exception->getMessage());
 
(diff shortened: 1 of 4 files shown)


app/Listeners/ProcessFederationFailure.php
```diff
@@ -15,7 +15,12 @@ public function handle(FederationProcessingFailed $event): void
         $federationId = $event->federation->id;
         $jobUuid      = $event->jobUuid;
 
-        $event->federation->update(['is_running' => 0]);
+        $event->federation->update([
+            'is_running' => 0,
+            'error' => true,
+            'error_text' => substr($event->exception->getMessage(), 0, 200),
```
calmacx, line 21: the 200 limit is a bit random

calmacx: ah just re-read the description and for whatever reason in the DB it's varchar(200)

loki-sinclair-hdruk (approved):

calmacx (approved):

app/Listeners/ProcessFederationFailure.php
loki-sinclair-hdruk: It was never meant to take the full stack trace. More a "something bad happened" note. But the core of this will capture the exception at least.

PR #1710, opened 2026-07-22
feat(GAT-9231): Add federation run-history API endpoint

equinoxmatt:
## Screenshots (if relevant)

image

## Describe your changes

Adds a read-only API for a federation's past run history, so the
frontend can show execution status and failure detail without
querying federation_job_runs directly. Three commits:

  - Extract the per-execution dataset fetch/parse logic (already used
    by the outcome email) onto FederationJobRun, so the new endpoint
    and the email share one implementation instead of two.
  - Adopt a Request::jwtUser() macro in FederationController, in place
    of the repeated $request->all()['jwt_user'] pattern, scoped to
    this controller as a candidate for wider rollout.
  - Add GET /teams/{teamId}/federations/{federationId}/history: a
    paginated list of past executions with status and per-dataset
    failure detail.
    
 Example response:
 
 ```
 {
    "current_page": 1,
    "data": [
        {
            "job_uuid": "",
            "started_at": "2026-07-15 08:42:59",
            "finished_at": "2026-07-15 08:42:59",
            "status": "failed",
            "message": "GWDM/2.0: must have required property 'gatewayId'",
            "failed_datasets": [
                {
                    "pid": "mock-dataset-bad",
                    "message": "GWDM/2.0: must have required property 'gatewayId'"
                }
            ]
        },
        {
            "job_uuid": "77",
            "started_at": "2026-07-15 08:18:34",
            "finished_at": "2026-07-15 08:18:34",
            "status": "success",
            "message": null,
            "failed_datasets": []
        }
    ],
    "first_page_url": "http://localhost:8000/api/v1/teams/5/federations/4/history?page=1",
    "from": 1,
    "last_page": 1,
    "last_page_url": "http://localhost:8000/api/v1/teams/5/federations/4/history?page=1",
    "links": [
        {
            "url": null,
            "label": "&laquo; Previous",
            "page": null,
            "active": false
        },
        {
            "url": "http://localhost:8000/api/v1/teams/5/federations/4/history?page=1",
            "label": "1",
            "page": 1,
            "active": true
        },
        {
            "url": null,
            "label": "Next &raquo;",
            "page": null,
            "active": false
        }
    ],
    "next_page_url": null,
    "path": "http://localhost:8000/api/v1/teams/5/federations/4/history",
    "per_page": 25,
    "prev_page_url": null,
    "to": 2,
    "total": 2
}
```

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-9231

- [x] I have added audit logs for new operation logic (if applicable)

### DIFF
--- a/app/Http/Requests/Federation/GetFederationHistory.php
+++ b/app/Http/Requests/Federation/GetFederationHistory.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace App\Http\Requests\Federation;
+
+use App\Http\Requests\BaseFormRequest;
+use App\Models\TeamHasFederation;
+
+class GetFederationHistory extends BaseFormRequest
+{
+    /**
+     *
+     * @return array<string, \Illuminate\Contracts\Validation\ValidationRule|array|string>
+     */
+    public function rules(): array
+    {
+        return [
+            'team_id' => [
+                'required',
+                'int',
+                'exists:teams,id',
+            ],
+            'federation_id' => [
+                'required',
+                'int',
+                'exists:federations,id',
+                function ($attribute, $value, $fail) {
+                    $exists = TeamHasFederation::where('federation_id', $value)
+                    ->where('team_id', $this->team_id)
+                        ->exists();
+
+                    if (!$exists) {
+                        $fail('The selected federation is not part of the specified team.');
+                    }
+                },
+            ],
+            'per_page' => [
+                'nullable',
+                'integer',
+                'min:1',
+            ],
+        ];
+    }
+
+    /**
+     * Add Route parameters to the FormRequest.
+     *
+     * @return void
+     */
+    protected function prepareForValidation()
+    {
+        $this->merge([
+            'team_id' => $this->route('teamId'),
+            'federation_id' => $this->route('federationId'),
+        ]);
+    }
+}
--- a/app/Models/FederationJobRun.php
+++ b/app/Models/FederationJobRun.php
@@ -5,8 +5,13 @@ namespace App\Models;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Prunable;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Support\Collection;
 
+/**
+ * @property-read string|null $started_at only populated on rows returned by executionsForFederation()
+ * @property-read string|null $finished_at only populated on rows returned by executionsForFederation()
+ */
 class FederationJobRun extends Model
 {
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: app/Http/Requests/Federation/GetFederationHistory.php [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+                'required',
+                'int',
+                'exists:federations,id',
+                function ($attribute, $value, $fail) {
+                    $exists = TeamHasFederation::where('federation_id', $value)
+                    ->where('team_id', $this->team_id)
+                        ->exists();
+
+                    if (!$exists) {
+                        $fail('The selected federation is not part of the specified team.');
```
loki-sinclair-hdruk, line 32: I'd opt to do a not found here, rather than provide crumbs.

equinoxmatt: I used the same as all the other federation request objects.  They all do it, even the `RunNowFedaration`.  So I opted for consistency.

Do you feel particularly strongly about this, or shall I keep it as it is?

app/Models/FederationJobRun.php
```diff
@@ -38,4 +44,75 @@ class FederationJobRun extends Model
...
+    public static function latestPerPidForExecution(int $federationId, string $jobUuid): Collection
+    {
+        return static::where('federation_id', $federationId)
+            ->where('job_uuid', $jobUuid)
+            ->whereIn('id', function ($query) use ($federationId, $jobUuid) {
+                $query->selectRaw('MAX(id)')
+                    ->from((new self())->getTable())
+                    ->where('federation_id', $federationId)
+                    ->where('job_uuid', $jobUuid)
+                    ->groupBy('pid');
```
loki-sinclair-hdruk, line 62: What's performance like on this? A quick scan would imply that this aggregates the same rows (potentially). While nothing wrong with it, I'd expect that to be backed by a composite, so that we're not doing extra work on a potentially large table. One to keep a watch on.

For others, a composite of: `(federation_id, job_uuid, pid, id)` would reduce the work required for both the outer filtering and the group by. This would be an exponential problem, as on a small table, it should be fine, but as it grows (as federation logs will) it'll get much slower over time.

equinoxmatt: Agreed, added some indexes, please take a look at the migration I just added.

app/Http/Requests/Federation/GetFederationHistory.php
loki-sinclair-hdruk: Nah, I too like consistency. I'll refactor later, as this leaves a trail if discovered.

loki-sinclair-hdruk (approved):

PR #1715, opened 2026-07-27
feat(GAT-2905): Add last run at to federations

equinoxmatt:
## Describe your changes

1. Turns on phpstan checkModelProperties so we get better types for models

2. Adds last_run_at timestamp to federations response on the API. This is required for some FE work but is also just useful information to have.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-2905

### DIFF
--- a/app/Models/FederationJobRun.php
+++ b/app/Models/FederationJobRun.php
@@ -83,6 +83,20 @@ class FederationJobRun extends Model
     }
 
     /**
+     * @param non-empty-array<int, int> $federationIds
+     * @return Collection<int, string> federation_id => last_run_at
+     */
+    public static function latestRunTimesForFederationIds(array $federationIds): Collection
+    {
+        return static::select('federation_id')
+            ->selectRaw('MAX(created_at) as last_run_at')
+            ->whereIn('federation_id', $federationIds)
+            ->groupBy('federation_id')
+            ->get()
+            ->pluck('last_run_at', 'federation_id');
+    }
+
+    /**
      * @return array<int, array{schema: ?string, message: string}>
      */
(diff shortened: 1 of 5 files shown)


app/Models/FederationJobRun.php
```diff
@@ -82,6 +82,20 @@ public static function executionsForFederation(int $federationId, int $perPage):
...
     }
 
+    /**
+     * @param non-empty-array<int, int> $federationIds
+     * @return Collection<int, string> federation_id => last_run_at
+     */
+    public static function latestRunTimesForFederationIds(array $federationIds): Collection
+    {
+        return static::select('federation_id')
+            ->selectRaw('MAX(created_at) as last_run_at')
```
calmacx, line 92: i forget how federations work but double check its not updated_at incase federation jobs can be re-ran?

phpstan.neon
```diff
@@ -8,6 +8,8 @@ parameters:
     # Level 9 is the highest level
     level: 3
 
+    checkModelProperties: true
```
calmacx, line 11: what does this do?

calmacx (approved):

phpstan.neon
equinoxmatt: Adds type inference for Eloquent models, rather than treating everything has mixed.

app/Models/FederationJobRun.php
equinoxmatt: It is coming from the `federation_job_runs` table. The created at will be when the job was last run, updated could be the result or some kind of error. For when it was run I think created_at is reliable.

PR #1719, opened 2026-07-28
feat(GAT-9143): #4 - Tweaks queue slightly to avoid job clashes causi…

loki-sinclair-hdruk:
…ng indexing to wait for some 5minutes before starting

### DIFF
--- a/app/Models/DatasetVersion.php
+++ b/app/Models/DatasetVersion.php
@@ -331,9 +331,17 @@ class DatasetVersion extends BaseTypesenseModel
         ]);
     }
 
+    /**
+     * @param  Builder<DatasetVersion>  $query
+     * @return Builder<DatasetVersion>
+     */
     public function makeAllSearchableUsing(Builder $query): Builder
     {
-        return $query->with(['spatialCoverage', 'dataset']);
+        // Pre-filter to indexable rows at the DB level (single correlated
+        // subquery) instead of pulling every historical version through the
+        // chunk and eager-loads only to be discarded by shouldBeSearchable()'s
+        // per-row query. See scopeIndexEligible() for the equivalence.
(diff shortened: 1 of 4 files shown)


soringumeni1 (approved):

app/Models/DatasetVersion.php
```diff
@@ -331,9 +331,17 @@ public function toSearchableArray(): array
...
+     * @return Builder<DatasetVersion>
+     */
     public function makeAllSearchableUsing(Builder $query): Builder
     {
-        return $query->with(['spatialCoverage', 'dataset']);
+        // Pre-filter to indexable rows at the DB level (single correlated
+        // subquery) instead of pulling every historical version through the
+        // chunk and eager-loads only to be discarded by shouldBeSearchable()'s
+        // per-row query. See scopeIndexEligible() for the equivalence.
+        return $query->indexEligible()->with(['spatialCoverage', 'dataset']);
```
calmacx, line 344: I know this isnt the code from the PR  - this `with(['spatialCoverage', 'dataset'])` is to make sure the parent dataset is loaded and the spatialCoverage which is an SQL table rather than in the JSON blob is loaded?

Feels like this is more suitable in the GwdmHandler in some sort of way to make sure these are (always?) loaded?

---> more making a note to myself to look into this

calmacx (approved):

app/Models/DatasetVersion.php
loki-sinclair-hdruk: Aye. I believe that the geolocation filter may play a part in this, and as it stands it needs to be computed before indexing, given it's not inside the model ordinarily. I imagined you'd pick up on this and add it to your existing pattern.

PR #1730, opened 2026-07-29
feat(GAT-8985): Adds a dataset title to federation failure email notification

equinoxmatt:
## Describe your changes

Adds a dataset title to federation failure email notification

### DIFF
--- a/app/Models/Dataset.php
+++ b/app/Models/Dataset.php
@@ -296,6 +296,27 @@ class Dataset extends Model
     }
 
     /**
+     * @param array<int, string> $pids
+     * @return array<string, string|null>
+     * 
+     */
+    public static function titlesForPids(array $pids): array
+    {
+        if (empty($pids)) {
+            return [];
+        }
+
+        $found = static::whereIn('pid', $pids)
+            ->with(['latestMetadata' => fn ($q) => $q->selectRaw('dataset_versions.id, dataset_versions.dataset_id, dataset_versions.title')])
+            ->get(['id', 'pid'])
+            ->mapWithKeys(fn ($dataset) => [
+                $dataset->pid => filled($dataset->latestMetadata?->title) ? $dataset->latestMetadata->title : null,
+            ]);
+
+        return collect($pids)->mapWithKeys(fn ($pid) => [$pid => $found[$pid] ?? null])->all();
+    }
+
+    /**
      * Helper function to grab dataset title from latest metadata.
      */
     public function getTitle(): string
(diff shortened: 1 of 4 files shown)


loki-sinclair-hdruk:
Looks sensible, but also looks to only cater for datasets we already know about? What happens if a new federation is run for the first time and errors?

equinoxmatt:
> Looks sensible, but also looks to only cater for datasets we already know about? What happens if a new federation is run for the first time and errors?

It falls back to just showing the PID. I think what you are asking is covered in this test: https://github.com/HDRUK/gateway-api/pull/1730/changes#diff-70822f4113d99001a7597eeba7e2e89d7ef0b8ae782299bda00ecc7b6bb40957R538

loki-sinclair-hdruk (approved):

app/Models/Dataset.php
```diff
@@ -295,6 +295,27 @@ public function searchByTitle(string $title): DatasetVersion|null
...
+    {
+        if (empty($pids)) {
+            return [];
+        }
+
+        $found = static::whereIn('pid', $pids)
+            ->with(['latestMetadata' => fn ($q) => $q->selectRaw('dataset_versions.id, dataset_versions.dataset_id, dataset_versions.title')])
+            ->get(['id', 'pid'])
+            ->mapWithKeys(fn ($dataset) => [
+                $dataset->pid => filled($dataset->latestMetadata?->title) ? $dataset->latestMetadata->title : null,
```
calmacx, line 313: Just FYI - there is a great deal of inconsistency across the repo with this - there is also a `short_title` that is often used. And also a fallback to `latestMetadata->metadata['summary']['title']` which is technically the source of truth. 

You could fallback to these if the title is unfilled 

I would suggest moving this to the GwdmHandler, however:
1) titles are unlikely to move in different GWDM version summary fields
2) we're often relying on the `$dataset->latestMetadata->title` in many other places already anyway

^^ leave it up to you if you want to look into understanding and/or changing this

calmacx (approved):

app/Models/Dataset.php
equinoxmatt: @calmacx thanks useful info. 

When you say move to the GwdmHandler, do you mean something like a resolveTitle()/getTitle() method on it that encapsulates the full fallback, or did you have something else in mind?

I agree I would not want to add fallback code in here.

calmacx: Yeah exactly something like that. As we know, this is a "problem" all over the place, so happy to leave to me as part of a ticket I have that is simply 'remove hard coding of GWDM paths' which this falls into :)

PR #1749, opened 2026-08-06
feat(GAT-8899): improves call speeds to return teams

loki-sinclair-hdruk:
## Describe your changes

Upon investigation this code was really poor. I discovered the following:
- For every user of every team `roles.permissions` is called inside a foreach. That means, for a full page (25) of teams with 10 users, you're looking at 250 queries per.
- For every team, we'd fetch `TeamHasNotifications`, then for every record, pull `Notification`, which essentially exposed an N+1 inside an N+1. Fun!
- We then, for reasons unknown to me, throw away an eagerly loaded `TeamHasAlias` and pull it again manually?! Bonkers.

Optimisations made:
- Batched the per-team/per-user N+1 queries (roles, notifications and aliases) into `whereIn` lookups.
- Added an Octane safe (no stale, or drifting static states) instance cache, so that the `$appends` accessors don't fire DB queries per user when many users are serialised at once.
- Preload before serialising teams' users, in index, show and listOfAdmin functions.

Optimisations thus lead to use using existing eagerly loaded relations. In profiling, db calls went from ~1315 to ~15 per.

## Issue ticket link
https://hdruk.atlassian.net/browse/GAT-8899

None

No

### DIFF
--- a/app/Http/Traits/TeamTransformation.php
+++ b/app/Http/Traits/TeamTransformation.php
@@ -10,6 +10,7 @@ use App\Models\Notification;
 use App\Models\TeamHasAlias;
 use App\Models\TeamHasNotification;
 use App\Models\TeamUserHasNotification;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 
 trait TeamTransformation
 {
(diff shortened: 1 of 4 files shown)


app/Http/Traits/TeamTransformation.php
```diff
@@ -101,21 +134,22 @@ public function getTeams(array $teams): array
...
             $tmpTeam['users'] = $tmpUser;
 
-            $notifications = TeamHasNotification::where('team_id', $tmpTeam['id'])->get()->toArray();
             $tmpNotification = [];
-            foreach ($notifications as $value) {
-                $notification = Notification::where('id', $value['notification_id'])->firstOrFail();
+            foreach ($notificationsByTeamId->get($tmpTeam['id'], collect()) as $value) {
+                $notification = $notificationsById->get($value['notification_id']);
+                if ($notification === null) {
+                    throw (new ModelNotFoundException())->setModel(Notification::class, [$value['notification_id']]);
```
loki-sinclair-hdruk, line 141: Potentially risky change here. But seeing how many entities demand a Notification model, it seemed silly that we'd not throw when one was needed and not found.

calmacx (approved):

PR #1762, opened 2026-08-12
feat(GAT-9459): Add RENEWING status with access/renewal eligibility checks

equinoxmatt:
## Describe your changes

Users can currently only reapply for CDS access after it's already expired. This adds RENEWING as a real status so a renewal can be in progress without ever touching a user's existing access or expiry.

Also adds a new hasAccess check which acts as a computed property, derived from status and expiry.

## Issue ticket link

https://hdruk.atlassian.net/browse/GAT-9459

### DIFF
--- a/app/Models/CohortRequest.php
+++ b/app/Models/CohortRequest.php
@@ -58,6 +58,18 @@ class CohortRequest extends Model
     public const REQUEST_PENDING = 'PENDING';
     public const REQUEST_EXPIRED = 'EXPIRED';
 
+    /**
+     * Statuses under which a user currently has CDS access.
+     */
+    public const ACCESS_GRANTING_STATUSES = [
+        CohortRequestStatus::APPROVED,
+        CohortRequestStatus::RENEWING,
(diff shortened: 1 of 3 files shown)


app/Models/CohortRequest.php
```diff
@@ -85,6 +97,30 @@ public function calculateTrueExpiry(string $expiryField = 'request_expire_at'):
         return Carbon::instance(min($candidates));
     }
 
+    public static function hasAccess(CohortRequest $request): bool
```
calmacx, line 100: wouldn't you check `CohortRequest::...(['request_status' => 'APPROVED'])->first();` to check if someone has access?

maybe this is a safer check on the expiry date rather than the request_status? theres probably a lot of inconsistencies in the code already

equinoxmatt: This is purely based on a already retrieved record. Checks will be added/adjusted later for using the new logic.

calmacx (approved):

